Issue #2, opened 2026-09-17
Activation feedback: microphone light and acknowledgement sound


jerryfane:
Parent: #1

## Goal

Give immediate, local feedback when the exact wake phrase is accepted, before any command reaches the agent.

## Behavior

- On **“Hey Voice”**, switch the supported microphone indicator into a clear listening state.
- Play a short, pleasant acknowledgement sound (“earcon”) through the configured output.
- Keep the indicator active while capturing/processing the accepted request.
- Restore the normal indicator state after completion, timeout, cancellation, or error.
- Use distinct, subtle feedback for success, failure, and timeout only if those sounds remain configurable and non-disruptive.

## Implementation constraints

- Feedback is emitted by the local wake/session layer; it must not call the LLM.
- Add a small indicator interface rather than coupling the session loop directly to the Anker PowerConf HID implementation.
- Reuse the configured `telephony_hid` device where the PowerConf supports an indicator report; fail safely if the device or report is unsupported.
- Make acknowledgement sound enable/disable, volume, and sound selection configurable.
- Prefer a tiny generated PCM waveform or an explicitly licensed bundled asset.
- Do not let playback feed back into wake recognition; serialize or suppress capture as required.
- `voice off` must leave the indicator in the off/idle state and play no wake sound.

## Acceptance criteria

- [ ] Exact `Hey Voice` produces visible microphone feedback within 150 ms of wake acceptance.
- [ ] A short acknowledgement sound plays once per accepted wake.
- [ ] Random noise, embedded phrases, and approximate wake phrases produce no light or sound.
- [ ] Indicator state is restored on success, failure, timeout, cancellation, and service shutdown.
- [ ] Unsupported HID hardware does not prevent Voice from accepting commands.
- [ ] Behavior is covered with fake indicator/player tests and smoke-tested on the installed PowerConf.
- [ ] README and example configuration document feedback settings.

jerryfane:
## Status: implemented, reviewed, on-device criterion blocked on the owner

Implementation is in PR #8, approved at head `a69dcc7` by an independent review-only job (`local-review-review-router-18d627cc59b69d43-1`), with one later P3 answered at `72cdec6`.

Acceptance criteria:

| Criterion | State |
| --- | --- |
| Visible mic feedback within 150 ms of wake acceptance | Implemented; the light is set before the sound plays, asserted by ordering test. **Not confirmed on hardware.** |
[...]

PR #8, opened 2026-09-17
Acknowledge accepted wake phrases with a light and a sound

jerryfane:
Partially addresses #2 — the on-device smoke-test criterion stays open (blocked by #9); #2 must not auto-close on merge.

## What changed

Voice now acknowledges an accepted wake phrase locally, before anything reaches the agent: the speakerphone indicator switches to a listening state and a short generated tone plays once. No model call, no network.

- `internal/feedback` — `Indicator` interface, `Notifier` (light first, then sound), `TelephonyLight`, and `Nop` for unsupported hardware.
- `internal/hid` — parses the device's HID report descriptor to discover which LED usages it actually advertises, and owns the hidraw node so writes retain the whole output report.
- `internal/audio/earcon.go` — generated waveforms (`chime`, `blip`, `two-up`), raised-cosine edges, no bundled asset and no licence question.
- `internal/session` — feedback fires only after `wake.Match` accepts, and the indicator is restored on every exit path.
- `feedback.{light,sound,volume}` config, validated at load; `voice doctor` reports what was selected and why.

## Why the off-hook LED is not the listening light

Capture holds the telephony off-hook bit for the entire session, otherwise the PowerConf streams digital silence. That bit lives in the same output report as the LEDs (report 2 on the installed device), so an indicator that wrote its own report would have cleared off-hook and silenced the microphone. `hid.Telephony` therefore retains the full report across writers, and `feedback.light` selects among the other indicators (`auto`, `mic`, `ring`, `mute`, `hold`, `none`).

## Acceptance criteria

| Criterion | State |
| --- | --- |
| Visible mic feedback on exact `Hey Voice` | Code path verified locally; **LED not confirmed on hardware** (see below) |
| One acknowledgement sound per accepted wake | Verified end-to-end: one 180 ms 16 kHz mono earcon, peak 14745 at volume 0.5, edges at 0 |
| Noise, embedded phrases and approximate phrases stay dark and silent | Verified end-to-end and in `internal/session` tests |
| Indicator restored on success, failure, timeout, cancellation, shutdown | `internal/session` tests pin the transitions per command |
| Unsupported HID hardware does not block commands | Verified: unreadable descriptor disables the light, `voice doctor` explains it, commands still run |
| Fake indicator/player tests | `internal/feedback`, `internal/session`, `internal/hid`, `internal/audio` |
| README and example config document the settings | Done |
| Smoke-tested on the installed PowerConf | **Blocked** |

## Verification

Local, with the built binary and a fake capture/STT/brain config:

- accepted wake (`hey voice turn on the light`) produced exactly one earcon WAV: 180 ms, 16 kHz mono, peak 14745 (= 0.5 volume x 0.9 x full scale), first and last samples 0;
- non-wake transcript (`the television is loud`) produced no audio output at all;
- unparseable descriptor path: `wake feedback INFO no indicator (no supported device configured); 180 ms sound`, and the command still ran.

`go test ./...`, `go vet ./...`, `gofmt -l .` clean; `CGO_ENABLED=0 GOOS=linux GOARCH=arm64` build ok. The session and HID tests were mutation-checked: firing feedback before the wake gate, and dropping the per-command restore, both fail the suite.

## Not verified

The on-device smoke test is blocked, and not by this change. The pi-burj seat found no `heard=`/`peak=` telemetry anywhere in seven days of `journalctl -u voice`, and `arecord` on the PowerConf under the unit's real uid/gid/supplementary groups measured exactly `peak_abs=0`, including while a 1 kHz tone was playing through the same device. Production capture is therefore unproven; until a human confirms the microphone hears anything, an LED/earcon smoke test on that device would prove nothing. The LED usage mapping in the descriptor test fixture also still needs the lossless descriptor dump from the device.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d6269f619d09f4-1`

**Summary:** Reviewed jerryfane/voice PR #8 at exact head 2625ed386df7191519e3ebb38dd4d87df423f4d9 (2 commits: 87a938c feature, 2625ed3 real descriptor fixture + bounds) against base 126042d. Build/vet/gofmt/full test suite/independent 30s fuzz run all pass; descriptor.go's bit-offset accounting was manually traced against the real PowerConf fixture and matches exactly. The off-hook bit is architecturally protected from indicator writes (leds map in light.go deliberately excludes LEDOffHook), which correctly forecloses the requester's headline 'digitally silent' scenario for the normal (descriptor-parses) path. However, the descriptor-unparseable fallback used precisely when a device is malformed or sysfs is unreadable (the exact case this PR's second commit targets) writes a report one byte shorter than the real device's report-2 layout and shorter than the pre-PR code's unconditional write, with zero test coverage — a genuine regression in a documented resilience path, requesting changes on that basis. Two secondary, lower-severity issues also found. Disclosure requested by operator: this review was performed by a Claude runtime, same model family as the implementing seat (only Claude had credentials on this host).
**Severity:** `P1`

**Findings**
- **Descriptor-unparseable fallback writes an undersized HID report, regressing the documented sysfs-unreadable resilience path** (P1)
  - detail: hid.Open() leaves t.outs nil whenever ReadDescriptor fails (sysfs unreadable) or ParseOutputs returns one of its nil,err paths (empty/truncated/implausible descriptor -- exactly the malformed-descriptor class this PR's bounds now reject cleanly instead of hanging). command.go's setOffHook then falls back to SetReportBit(Bit{ReportID:2,Byte:0,Mask:1}), which goes through payload(): PayloadBytes(2) on a nil Outputs returns 0 (nil-safe), so payload() takes the 'want=bit.Byte+1' branch and grows the retained state to 1 byte instead of the real device's 2-byte payload. The resulting write is 2 bytes total (ID+1 byte) versus the pre-PR code's 3 bytes (ID+2 bytes) and versus what SetReportBit's own doc comment ('the standard telephony report') and command.go's comment ('the standard headset report layout') claim to reproduce. Many USB HID devices validate Set_Report length against the descriptor's declared size; a short write can be rejected (making capture fail entirely, since setOffHook's error aborts Stream() before any audio is read) or leave firmware state undefined, in precisely the scenario Open()'s own doc comment says should 'keep working on hosts where sysfs is not reachable.' No test constructs a Telephony with nil outs and calls SetReportBit/Open to catch this -- confirmed via grep across internal/hid and internal/audio test files, zero references.
  - evidence_locator: Static trace, cross-checked against git diff 126042d..2625ed3 -- internal/audio/command.go and internal/hid/telephony_test.go:64; no existing test exercises this path (grep -rn 'SetReportBit|hid.Open' internal/**/*_test.go).
  - id: F1
  - locator: internal/hid/telephony.go:38-49 (Open leaves t.outs nil on ReadDescriptor/ParseOutputs failure), internal/hid/telephony.go:145-160 (payload() sizes only from the touched bit when outs is nil: want=0 -> want=bit.Byte+1=1), internal/audio/command.go:61-66 (SetReportBit fallback call site, comment claims 'standard headset report layout'); compare pre-PR internal/audio/command.go@126042d which unconditionally wrote f.Write([]byte{2, v, 0}) (3 bytes: ID+2 payload); real device layout confirmed 2-byte payload for report 2 in internal/hid/telephony_test.go:64 (outs.PayloadBytes(2) == 2).
- **Telephony.write() releases the mutex before the physical hidraw write, allowing out-of-order I/O completion between concurrent writers** (P2)
  - detail: write() locks, mutates the retained per-report state in place, builds a report snapshot, then unlocks before doing the actual os.OpenFile+Write. Two concurrent callers (the capture goroutine's one-time setOffHook at Stream() start/end, internal/audio/command.go:72-83, versus the session goroutine's feedback light writes, internal/feedback/light.go:98-99) can have their physical writes land on the device out of order relative to their lock-acquisition order, so a stale, less-complete snapshot can overwrite a newer one on the wire. Because the off-hook usage is excluded from the indicator's leds map (internal/feedback/light.go:31-39), this cannot by itself clear the off-hook bit mid-session (every snapshot taken after the initial off-hook mutation still carries it set), so the catastrophic 'mic goes silent' scenario is not reachable through this race in the current call pattern. The realistic consequence is a transient or stuck-stale LED bit at session start/shutdown when the capture goroutine's one-shot off-hook set/clear races the feedback goroutine's writes to the same report. Recommend holding the lock across the I/O, or serializing writes through a single-writer goroutine/channel, to make 'last logical mutation wins on the wire' an invariant rather than a probabilistic outcome.
  - evidence_locator: Static analysis of internal/hid/telephony.go:121-143; go test -race could not be run in this sandbox (cgo blocked, see tests_run) so this is not race-detector-confirmed, only traced.
  - id: F2
  - locator: internal/hid/telephony.go:121-143 (lock covers only the state mutation and snapshot build at lines 122-132; os.OpenFile/Write at lines 134-141 run unlocked)
- **Push/Pop items restore only the usage page, not the full HID global item state** (P3)
  - detail: Per the HID spec, Push (0xA4) and Pop (0xB4) snapshot/restore the entire global item state (usage page, report ID, report size, report count, logical/physical bounds, unit, etc.), not just usage page. This parser only tracks usage page across push/pop. The real PowerConf fixture never uses push/pop (verified by manual byte-trace), so this is unexercised by any test and causes no observed misbehavior today, but a future device whose descriptor pushes state, changes report size/count/ID inside a nested collection, and pops expecting restoration would be misparsed (wrong bit offsets or wrong report ID association) without any error being raised.
  - evidence_locator: Manual trace of internal/hid/telephony_test.go's powerConfDescriptor bytes (no 0xA4/0xB4 bytes present); code reading of internal/hid/descriptor.go:190-196.
  - id: F3
  - locator: internal/hid/descriptor.go:190-196 (tagPush/tagPop only save/restore `page`)

**Tests Run**
- CGO_ENABLED=0 go build ./... -> exit 0, clean (plain `go build ./...` fails in this sandbox with cgo blocked: 'fatal error: /usr/include/stdc-predef.h: Permission denied', unrelated to the PR)
- CGO_ENABLED=0 go vet ./... -> exit 0, clean
- gofmt -l . -> exit 0, no files listed (already formatted)
- CGO_ENABLED=0 go test ./... -> all packages ok (audio 0.014s, device 0.131s, feedback 0.015s, hid 0.012s, session 0.009s, vad 0.007s, wake 0.008s; brain/config/proc/speech/cmd/voice have no test files)
- CGO_ENABLED=0 go test ./internal/hid/... -run=FuzzParseOutputs -fuzz=FuzzParseOutputs -fuzztime=30s -> PASS, 4,355,082 execs in ~31s, 140 interesting corpus entries found, 0 crashes (independent of and exceeding the author's cited 45s/4.2M-exec run)
- CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o voice-arm64 ./cmd/voice -> exit 0; `file` confirms 'ELF 64-bit LSB executable, ARM aarch64 ... statically linked'
- go test -race ./internal/hid/... ./internal/feedback/... ./internal/session/... -> build failed: cgo required for -race, blocked by sandbox ('/usr/include/stdc-predef.h: Permission denied'); race detector could not be run, F2 is static-only
- Manual byte-by-byte decode of the powerConfDescriptor fixture (internal/hid/telephony_test.go:13-30) against ParseOutputs's algorithm, confirming all five LED bit positions (report 2, byte 0, masks 0x01..0x10) and the 2-byte payload size match the test's assertions exactly
- Manual decode of the fuzz-found crasher bytes in internal/hid/testdata/fuzz/FuzzParseOutputs/33f013f4578acdf2, confirming the encoded report count is 0x91283097 (~2.4B) and that the current maxFieldCount=1024 bound rejects it before any loop runs
- Independent arithmetic check of internal/audio/earcon.go's render(): confirmed the default 'chime' earcon is exactly 0.07s+0.11s=180ms at 16kHz, peak amplitude at volume=0.5 is 0.5*1*0.9*32767=14745 (int16), and both first/last samples are exactly 0 (env=0.5-0.5*cos(0)=0 at the boundary) -- matching the author's cited end-to-end figures without needing to reconstruct their CLI harness
- grep -rn 'SetReportBit|hid.Open' across internal/hid and internal/audio test files -> zero test references, confirming F1's fallback path is untested
- git status --porcelain / git diff --stat after all runs -> clean, no repository files modified by this review (fuzzing wrote no new corpus entries in this sandbox)

jerryfane:
## Independent review verdict, and answers to it

**Published here** (this PR comment) per directive 173461; a local job row alone is not a findable record, and a workflow note alone would not sit with the code.

| Field | Value |
| --- | --- |
| Verdict | `changes_requested` |
| Reviewer | `voice-review-claude`, review-only (`--no-fix-target`, `--org-role voice`) |
| Job | `local-review-review-router-18d6269f619d09f4-1` |
| Head reviewed | `2625ed386df7191519e3ebb38dd4d87df423f4d9` |
| Evidence | `executed`, 12 checks, 3 findings (P1/P2/P3) |
| Model family | Same family as the implementation. Only the Claude runtime has credentials on this host: codex returns `401 Unauthorized`, kimi has no credential configured. Stated rather than left implicit. |
| Superseded job | `local-review-review-router-18d62657f49a5bae-1` at `87a938c`, cancelled by me with no recorded result after I pushed a second commit; it carries no findings forward. |

Answers are at head `a69dcc79f3d3043a3833bfb1a9625627ad211fe6`. The coordinator asked for the defect *class* rather than the incidents, so for each item I state whether a **new** malformed descriptor, written with no further code, is now safe.

### `#8-f1` (P1) — descriptor-unparseable fallback wrote an undersized report

Correct, and the most serious of the three: it made the resilience path this PR introduced worse than the code it replaced. `hid.Open` leaves `outs` nil when sysfs is unreadable or the descriptor does not parse; `setOffHook` then used a bare `Bit{ReportID: 2, Byte: 0, Mask: 1}`, and `payload()` sized the report from the touched bit, writing two bytes (`02 01`) where the device declares three (`02 01 00`). Firmware treats a short report as a different report, which is precisely how an off-hook write leaves a microphone gated.

Fixed structurally: `hid.Bit` now carries `Payload`, the report's full payload length. The parser stamps it from the descriptor once every item for that report has been seen, and `hid.StandardOffHook` states its own length of 2. No writer asks a second object how long a report is, so report length no longer depends on whether the descriptor happened to parse. `TestFallbackOffHookWritesFullStandardReport` builds a handle with no parsed descriptor and asserts `02 01 00`; reverting `payload()` to the old `t.outs.PayloadBytes()` source fails it.

**Safe for a new case: yes.** The length travels with the `Bit`, and every construction site must fill it.

### `#8-f2` (P2) — mutex released before the physical write

Correct. `write()` snapshotted under the lock and wrote outside it, so the capture goroutine's off-hook write and the session goroutine's indicator write could land on the device out of order, letting an older snapshot overwrite a newer one on the wire. The lock now spans the I/O. These writes are three bytes, a few times per command, so serialising them costs nothing worth measuring.

`TestConcurrentWritersNeverEmitATornReport` drives 200 concurrent LED writes against a held off-hook bit and asserts every landed report is a whole three-byte report 2 with the off-hook bit intact. The reviewer could not run the race detector (cgo blocked in that sandbox, correctly flagged as static-only), so I ran the full suite under it here: `CGO_ENABLED=1 go test -race ./...`, all green.

**Safe for a new case: yes**, for writers inside Voice. An independent process writing the same hidraw node remains outside this type's control, which no locking here can fix.

### `#8-f3` (P3) — Push/Pop restored only the usage page

Correct per the HID spec. Push and Pop snapshot the entire global item state; restoring only the usage page silently misparses a device that changes report id, size or count inside a pushed scope — wrong offsets or wrong report, with no error raised. The parser now pushes and pops the whole global state as one value.

`TestPushPopRestoresEveryGlobalItem` parses a descriptor that pushes, declares a four-byte output in report 7, pops, then declares one bit in report 2, and asserts the later output resolves against the restored state. The page-only mutant fails it.

**Safe for a new case: yes**, for the globals the parser tracks (usage page, report id, report size, report count). Logical and physical bounds and units are deliberately not tracked, because nothing single-bit depends on them; if that changes they join the same struct rather than becoming four more loose variables.

### The truncation class, which the reviewer did not raise

Directive 173461 also required that a short or malformed read be *detected explicitly*, not silently parsed or reconstructed from a decoded table — the failure that produced this PR's first fixture. Three changes:

1. `ReadDescriptor` asks the kernel for the descriptor length (`HIDIOCGRDESCSIZE`) and returns **no descriptor at all** on a mismatch. That is the only real authority: a 158-byte prefix of the installed unit's 191-byte descriptor parses cleanly, balanced collections and all, while describing a device that does not exist. A byte-count comparison catches that; no amount of self-consistency checking does.
2. Unbalanced collections are rejected, which catches cuts landing mid-collection — including the first dump's own cut region: prefixes at 81, 92 and 100 bytes are all rejected, asserted in `TestTruncatedDumpIsRejectedNotPartiallyParsed`.
3. `hid.Open` never retains a partially parsed map. Any parse error leaves the handle advertising nothing, so it falls back to `StandardOffHook` and reports no indicator, instead of driving bits from half a descriptor.

**Safe for a new case: yes.** Any truncation changes the byte count, and the kernel's length is checked before the bytes are parsed. Where the ioctl is unavailable the descriptor is used and reported as *unverified* rather than silently trusted.

### The report-count hang, bounded by the input rather than by a ceiling

The previous commit's `maxFieldCount = 1024` was a special case shaped by the crasher, which is exactly the objection. Replaced: iteration walks only the usages the descriptor actually declares (`named()` takes the floor of the declared count, the usage list and the usage range), the report's bit budget is checked against what hidraw can carry **before** any iteration, and the arithmetic is `uint64`, so no size-times-count product can wrap.

Adversarial check, all rejected, largest input 5604 bytes:

| Descriptor | Result | Time |
| --- | --- | --- |
| report count `0xffffffff` | rejected: 536870912 bytes exceeds 4096 | 12.6 us |
| report size and count both `0xffffffff` | rejected: 2.3e18 bytes exceeds 4096 | 1.7 us |
| full 16-bit usage range with max count | rejected | 1.0 us |
| 400 output items each claiming a full usage range | rejected | 0.9 us |

**Safe for a new case: yes.** No firmware-supplied number drives iteration or allocation; only the descriptor's own byte count does.

The fuzz seed stays committed at `internal/hid/testdata/fuzz/FuzzParseOutputs/33f013f4578acdf2`.

### Verification at `a69dcc7`

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean.
- `CGO_ENABLED=1 go test -race ./...` — all packages green.
- `go test ./internal/hid/ -fuzz FuzzParseOutputs -fuzztime 60s` — PASS, 5,698,740 execs, no crashes.
- `CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build ./cmd/voice` — ok.
- Each of the three fixes mutation-checked by reverting it in isolation and confirming the matching new test fails.

Still open and not caused by this PR: the on-device smoke test, blocked by #9. Merge is not requested; no deploy or release.

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d627cc59b69d43-1`

**Decision:** `approved`
**Summary:** All three prior findings (F1 undersized fallback report, F2 lock-before-I/O race, F3 partial push/pop) are resolved at a69dcc7, verified by build/vet/test and standalone reproductions of the specific mechanisms (multi-item payload stamping, same-report offset continuity, push/pop field restoration). One new bounded P3 (parser work disproportionate to descriptor length for a UsageMin/UsageMax range, contradicting the 'one usage per two bytes' doc comment) does not block merge.
**Severity:** `P3`

**Findings**
- **Descriptor-unparseable fallback writes an undersized HID report — RESOLVED** (P1)
  - continues_uid: jerryfane/voice#8-f1
  - detail: hid.Bit gained a Payload field; ParseOutputs stamps it onto every bit after the full descriptor walk (descriptor.go:238-244), using the final (max) o.sizes[reportID] rather than a per-item snapshot, so growth across multiple Output items sharing one report ID is captured correctly. StandardOffHook carries Payload:2. Telephony.payload() now sizes exclusively from bit.Payload, never from t.outs, so a nil-outs fallback no longer produces an undersized write. Verified: (1) repo test TestFallbackOffHookWritesFullStandardReport passes, writing 02 01 00 (3 bytes); (2) standalone reproduction against a copy of descriptor.go with a report that has a 1-bit named field followed by a 2-byte unnamed field in the same report ID correctly stamps Payload=3 for the earlier bit; (3) standalone reproduction of two 1-bit named fields in the same report/item run correctly advances the running bit offset (byte0 bit0 and byte0 bit1) rather than colliding. Only two Bit{} construction sites exist in production code (StandardOffHook and the stamped parser output); grep confirms no other site can leave Payload zero.
  - evidence_locator: go test ./internal/hid/... -run TestFallbackOffHookWritesFullStandardReport: PASS; standalone verification module (descriptor.go copy) TestScratchMultiItemPayloadStamp and TestScratchOffsetContinuesAcrossItemsSameReport: PASS, run outside the repo tree, no repo files modified.
  - file: internal/hid/telephony.go:192-209, internal/hid/descriptor.go:33-49,238-244
  - state: answered
- **Telephony.write() releasing the mutex before the physical write — RESOLVED (narrow residual latency risk noted)** (P2)
  - continues_uid: jerryfane/voice#8-f2
  - detail: write() now holds t.mu across os.OpenFile and Write (defer t.mu.Unlock() at line 170), so the out-of-order stale-snapshot-overwrite race is closed: 'last logical mutation wins on the wire' is now enforced by the lock rather than probabilistic. No reentrant call exists inside write() that could self-deadlock; Describe()/Has() read t.outs without locking, but t.outs is set once in Open() and never mutated afterward, so this is not a race. Residual concern: the feedback light's Set() (internal/feedback/light.go:98-99, called synchronously from the session loop's accepted() path, which the Indicator interface's own doc comment at feedback.go:34 requires to change 'within a perceptible fraction of a second') can block on t.mu if it lands while the capture goroutine's off-hook write is in flight. This window is narrow in practice: setOffHook is called only once at Stream() start and once (deferred) at shutdown (internal/audio/command.go:80,84), not per-utterance, so contention requires a wake phrase to be recognized in the same instant as session startup/shutdown, and the physical write itself is normally sub-millisecond. This is a different, lower-severity concern than the original data-corruption race, which is fixed.
  - evidence_locator: Code read of internal/hid/telephony.go:164-190, internal/feedback/light.go:46-107, internal/feedback/feedback.go:32-38, internal/audio/command.go:70-84, internal/session/session.go (Run/accepted control flow). go test -race could not run (CGO_ENABLED=1 go test -race ./internal/hid/... ./internal/audio/... ./internal/feedback/...: build failed, fatal error: /usr/include/stdc-predef.h: Permission denied — cgo blocked by sandbox); repo's new TestConcurrentWritersNeverEmitATornReport (25x4 concurrent LED togglers plus an off-hook set) passes and confirms the off-hook bit survives concurrent writers under CGO_ENABLED=0 go test ./internal/hid/....
  - file: internal/hid/telephony.go:164-190
  - state: answered
- **Push/Pop restoring only usage page — RESOLVED for the fields the parser actually uses** (P3)
  - continues_uid: jerryfane/voice#8-f3
  - detail: The globals struct now carries page, reportSize, reportCount, and reportID, and Push/Pop save/restore the whole struct rather than just page. These four fields are exactly what the parser consults anywhere in bit-address computation (report ID, byte offset via reportSize*reportCount running totals, byte/mask math); logical/physical bounds and units are never read by ParseOutputs, Bit, or Outputs, so their omission from Push/Pop cannot misparse a single-bit output's report ID, byte, or mask. Verified with the repo's new TestPushPopRestoresEveryGlobalItem (push inside a nested collection that changes report ID/size/count, then pop, then a subsequent single-bit output resolves against the restored report 2/size 1/count 1 state) and a standalone same-report multi-field check confirming correct offset continuity.
  - evidence_locator: go test ./internal/hid/... -run TestPushPopRestoresEveryGlobalItem: PASS; grep -rn 'logical\|physical\|unit' internal/hid/descriptor.go: no matches, confirming those fields are never consulted.
  - file: internal/hid/descriptor.go:128-146,226-231
  - state: answered
- **ParseOutputs accepts a 17-byte descriptor that costs up to ~35ms, contradicting the 'at most one usage per two bytes' bound claim** (P3)
  - detail: The comment at descriptor.go:108-113 claims iteration only ever walks usages the descriptor actually declares, 'at most one per two bytes.' This is false for UsageMin/UsageMax ranges: a 6-byte UsageMin(0)+UsageMax(0xFFFF) pair declares a 65536-wide span, and named()'s floor is by ReportCount (attacker-controlled, up to the byte-budget ceiling), not by descriptor bytes consumed. I built a 17-byte descriptor (UsagePage, ReportID=2, ReportSize=1, ReportCount=N, UsageMin=0, UsageMax=0xFFFF, Output) that ParseOutputs *accepts* (does not reject) and that costs ~240us at N=1000 up to ~10-35ms at N=32768 (the maxIndexedBits ceiling) -- several orders of magnitude more than the ~13us the operator measured for their four adversarial (rejected) descriptors, because this one is never rejected. The mitigating factor: maxIndexedBits=32768 is a hard, global, cumulative cap checked immediately before each Output item's indexing loop runs (descriptor.go:199-202), so total parser work across any single call is bounded at a fixed ~35ms regardless of descriptor size or how the budget is spread across report IDs -- I confirmed this by chaining 250 report IDs each attempting a 200-usage range (3752-byte descriptor): the parser rejects with 'report descriptor names more than 32768 single-bit outputs' in under 1ms, never running the loop that would exceed the cap. So this is bounded, not unbounded: a real but low-severity gap between the code's proportionality claim and its actual (still-safe) behavior, worth a comment fix but not a functional or security defect at the P1/P2 level.
  - evidence_locator: Standalone verification module (copy of descriptor.go, no repo files modified) run via go test -v: TestAmplificationRatio shows count=1000/desclen=17/elapsed=240.784us, count=8000/elapsed=1.902298ms, count=32768/elapsed=10.116ms (a second run of the same case hit 35.5ms); TestAdversarialManySmallReportsCumulativeCapped (250 report IDs, 3752-byte descriptor) rejects in 724.487us with the maxIndexedBits error, confirming the global cap prevents unbounded work.
  - file: internal/hid/descriptor.go:108-113,197-211
  - state: answered

**Tests Run**
- go version -> go1.26.4 linux/amd64
- CGO_ENABLED=0 go build ./... (workspace) -> success, no output
- CGO_ENABLED=0 go vet ./... (workspace) -> success, no output
- CGO_ENABLED=0 go test ./... (workspace) -> ok for internal/audio, internal/device, internal/feedback, internal/hid, internal/session, internal/vad, internal/wake
- gofmt -l . (workspace) -> no output (clean)
- CGO_ENABLED=1 go test -race ./internal/hid/... ./internal/audio/... ./internal/feedback/... -> build failed: fatal error: /usr/include/stdc-predef.h: Permission denied (cgo blocked by this sandbox; gap covered by the requester's own reported -race run at this head)
- Standalone verification module (copy of internal/hid/descriptor.go into a temp dir outside the repo, module hidcopy; repo left untouched) go test -v: TestScratchMultiItemPayloadStamp PASS, TestScratchOffsetContinuesAcrossItemsSameReport PASS, TestAdversarialHugeCountRejectedFast PASS (14.167us), TestAdversarialHugeRangeSmallCountBounded (accepted, 35.5ms -> new P3 finding), TestAdversarialManySmallReportsCumulativeCapped PASS (724.487us), TestAmplificationRatio (informational, 240.8us-10.1ms across N=1000/8000/32768)

jerryfane:
## Follow-up verdict: approved, plus one new P3 answered

Published here per directive 173461, same channel as the previous round.

| Field | Value |
| --- | --- |
| Verdict | `approved` |
| Reviewer | `voice-review-claude`, review-only (`--no-fix-target`, `--org-role voice`) |
| Job | `local-review-review-router-18d627cc59b69d43-1` |
| Head reviewed | `a69dcc79f3d3043a3833bfb1a9625627ad211fe6` |
| Prior findings | `#8-f1` (P1), `#8-f2` (P2), `#8-f3` (P3) — all resolved and `answered`, each verified by standalone reproduction of its mechanism, not by reading |
| New finding | `#8-f4` (P3) — answered below, fixed at `72cdec61647927f5250e6cc8197b189e95f3512b` |
| Model family | Same family as the implementation, again stated: codex returns `401 Unauthorized` and kimi has no credential on this host. |

### `#8-f4` (P3) — the stated bound was not the bound

The reviewer is right, and the finding is well aimed: it attacks the *claim* rather than the code. My comment said iteration walks at most one usage per two descriptor bytes. A six-byte usage-minimum/maximum pair declares a 65536-wide span, so that sentence was false, and a 17-byte descriptor (`UsagePage`, `ReportID`, `ReportSize 1`, `ReportCount N`, `UsageMin 0`, `UsageMax 0xffff`, `Output`) was *accepted* and cost tens of milliseconds at the ceiling instead of the microseconds I had measured on descriptors that were rejected. My four adversarial cases all tripped the byte-budget check, so none of them exercised the path where a range is cheap to declare and expensive to index.

Not a hang, and not a crash, but the guard was not length-derived for ranges, which is exactly the property directive 173461 asked for. Fixed by capping indexed single-bit outputs at eight per descriptor byte, so work stays proportional to the input whether usages are enumerated or declared as a range. Real devices keep a wide margin: the installed speakerphone names six outputs in 191 bytes.

`TestUsageRangeCannotOutrunDescriptorLength` pins both halves — the 17-byte range descriptor is rejected in under 5 ms, and a legitimate eleven-LED array declared as the range `0x17..0x21` still parses with the microphone LED landing at byte 1 mask `0x04`. Restoring the old fixed ceiling fails it.

**Safe for a new case: yes.** No firmware-supplied number, and no declared span, can drive more work than the descriptor's own length allows.

### `#8-f2` residual latency, acknowledged rather than restructured

The reviewer noted the light's `Set()` can now block on the mutex behind a capture-path off-hook write, against a doc comment promising the indicator changes within a perceptible fraction of a second. Judgement: leave it. Off-hook writes happen twice per service lifetime, at stream start and stop, and each is an `open`, a three-byte `write` and a `close` on a hidraw node — sub-millisecond, and not contended with the wake path in any realistic interleaving. Trading that for lock-free write ordering would reintroduce `#8-f2`, which was the more serious defect. If measurement ever shows otherwise, the fix is a single-writer goroutine owning the node, not a narrower lock.

### Verification at `72cdec6`

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean.
- `CGO_ENABLED=1 go test -race ./...` — all packages green.
- `go test -fuzz FuzzParseOutputs -fuzztime 30s` — PASS, seed corpus still committed.
- F4 mutation-checked: reverting the cap to the fixed ceiling fails the new test.
- End-to-end with the built binary at this head: one accepted wake produced exactly one 180 ms 16 kHz earcon, peak 14745, first and last samples 0; a non-wake transcript produced no audio; and the descriptor-unparseable fallback wrote the full three-byte report `02 00 00`, which is F1's fix visible in the real binary rather than only in a unit test.

### Status

Merge is not requested and I hold no merge authority here; the coordinator queues that decision for the owner. The on-device criterion of #2 stays **blocked on the owner**, not failed: the physical session is staged and frozen, and host-side causes for #9 are eliminated with a positive control while the device-mute hypothesis is untested.

jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d6297d58c6b2d6-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d6297d58c6b2d6-1`

**Decision:** `approved`
**Summary:** Reviewed the single delta commit 72cdec6 (touching only internal/hid/descriptor.go and descriptor_test.go) against prior-reviewed head a69dcc7 at PR #8 head 72cdec61647927f5250e6cc8197b189e95f3512b. F4 (amplification ratio) is resolved: the fixed 32768-bit ceiling was replaced with a length-proportional cap (8 indexed single-bit outputs per descriptor byte), checked as a running total across all Output items before the expensive per-field loop runs, so work is now bounded by input size for usage ranges as well as enumerated usages. Independently confirmed via build/vet/tests, a 20s fuzz run, and three standalone scratch-module checks (mutation test proving the fix is load-bearing, multi-item accumulation into one report, and a hand-traced legitimate 11-LED range) that the cap rejects the original 17-byte pathological descriptor in <1ms, cannot let named() exceed a report's remaining bit budget, accumulates correctly across multiple Output items to the same report, and does not falsely reject realistic telephony/LED descriptors (the repo's own 191-byte/6-output fixture sits at ~0.03 outputs/byte, far under the 8/byte ceiling). F1 (Payload stamping) and F2 (lock-across-I/O) are untouched by this commit (zero diff in telephony.go/telephony_test.go since a69dcc7) and remain resolved as previously verified. F3 (whole-globals push/pop) is untouched by this commit's diff and remains resolved. No new defects found in the changed files. Worktree left clean; no repo files were edited.

**Findings**
- **ParseOutputs' bit-indexing cap is now length-proportional (8 per descriptor byte), closing the amplification gap** (P3)
  - continues_uid: jerryfane/voice#8-f4
  - detail: Replaced fixed maxIndexedBits=32768 ceiling with indexedBitsPerByte=8 compared against len(desc)*8, checked as a running cumulative total across all Output items (not per-item, not per-report) before the O(fields) indexing loop executes. Confirmed the original 17-byte UsageMin(0)/UsageMax(0xffff) pathological descriptor is now rejected in well under 1ms rather than costing up to ~35ms. named() cannot return a span exceeding a report's remaining bit budget since its count argument is the already-budget-checked g.reportCount. The cap accumulates correctly across multiple Output items sharing a report ID (standalone test: 4x2001-field items in a 64-byte descriptor reject on the first item once indexed>512). A legitimate compact 11-usage LED range (14 bytes) still parses with correct byte/mask placement (off-hook byte0/mask0x01, mic byte1/mask0x04), matching the new test's own assertions, which I independently hand-verified rather than trusting. Real devices in Voice's target class (telephony/LED HID pages) are very unlikely to trip the 8/byte ratio since those pages have only dozens of defined usages total; the repo's real 191-byte/6-output fixture runs at ~0.03/byte. Mutation check: reverting to the old fixed ceiling in a standalone scratch copy makes the new TestUsageRangeCannotOutrunDescriptorLength fail (16-byte descriptor indexes 32768 usages unrejected), confirming the fix is load-bearing.
  - evidence_locator: CGO_ENABLED=0 go test ./internal/hid/... -v: all tests including TestUsageRangeCannotOutrunDescriptorLength PASS (0.012s total). CGO_ENABLED=0 go test ./internal/hid/... -fuzz FuzzParseOutputs -fuzztime 20s: 1.25M executions, no crashes. Standalone scratch module (full copy of internal/hid outside repo tree, deleted after use): mutation check confirms old ceiling fails the new test; TestCapAccumulatesAcrossOutputItemsSameReport and TestCapAllowsAccumulationUnderRatio (both written and run by me, not shipped in the repo) PASS, confirming cross-item accumulation and legitimate-range acceptance. No repo files modified (git status clean throughout).
  - file: internal/hid/descriptor.go:108-119,199-205
  - state: resolved
- **Descriptor-unparseable fallback writes an undersized HID report — unchanged since a69dcc7, remains resolved** (P1)
  - continues_uid: jerryfane/voice#8-f1
  - detail: git diff a69dcc7..72cdec6 -- internal/hid/telephony.go internal/hid/telephony_test.go is empty: this commit touches only descriptor.go/descriptor_test.go and does not alter the Payload-stamping logic, Bit struct, or StandardOffHook. The prior resolution stands unchanged.
  - evidence_locator: git diff a69dcc7 72cdec6 -- internal/hid/telephony.go internal/hid/telephony_test.go: no output (zero diff). CGO_ENABLED=0 go test ./internal/hid/... -run TestFallbackOffHookWritesFullStandardReport -v: PASS.
  - file: internal/hid/telephony.go:192-209, internal/hid/descriptor.go:33-49,238-244
  - state: resolved
- **Telephony.write() releasing the mutex before the physical write — unchanged since a69dcc7, remains resolved (same narrow residual latency note as before)** (P2)
  - continues_uid: jerryfane/voice#8-f2
  - detail: telephony.go is byte-for-byte unchanged in this commit. The lock-across-OpenFile-and-Write fix from a69dcc7 stands; the previously-noted narrow residual latency window (feedback light Set() can block on t.mu if it lands exactly when the capture goroutine's off-hook write is in flight) is unchanged and still low-severity given setOffHook only runs at Stream() start/shutdown, not per-utterance.
  - evidence_locator: git diff a69dcc7 72cdec6 -- internal/hid/telephony.go: empty. CGO_ENABLED=0 go test ./internal/hid/... -run TestConcurrentWritersNeverEmitATornReport -v: PASS. CGO_ENABLED=1 go test -race could not run in this sandbox (fatal error: /usr/include/stdc-predef.h: Permission denied), same limitation as the prior review.
  - file: internal/hid/telephony.go:164-190
  - state: resolved
- **Push/Pop restoring only usage page — unchanged since a69dcc7, remains resolved for the fields the parser uses** (P3)
  - continues_uid: jerryfane/voice#8-f3
  - detail: The globals struct (page, reportSize, reportCount, reportID) and the Push/Pop save/restore logic at descriptor.go:132-146 and 229-234 are byte-for-byte identical before and after this commit's diff, which only touched the const block/comment (108-119) and the indexing cap check (199-205). No regression introduced.
  - evidence_locator: git diff a69dcc7 72cdec6 -- internal/hid/descriptor.go shows only lines 108-119 and 199-205 changed; lines 128-146 and 226-231 (globals struct, Push/Pop) untouched. CGO_ENABLED=0 go test ./internal/hid/... -run TestPushPopRestoresEveryGlobalItem -v: PASS.
  - file: internal/hid/descriptor.go:128-146,226-231
  - state: resolved

**Tests Run**
- CGO_ENABLED=0 go build ./... (clean)
- CGO_ENABLED=0 go vet ./... (clean)
- gofmt -l ./internal/hid/ (clean, no output)
- CGO_ENABLED=0 go test ./internal/hid/... -v (all pass: TestParseOutputsRejectsImplausibleItemCounts, TestPushPopRestoresEveryGlobalItem, TestTruncatedDumpIsRejectedNotPartiallyParsed, TestUsageRangeCannotOutrunDescriptorLength, TestPowerConfLEDsDecodeToReportTwo, TestParseOutputsReportsDevicesWithNoOutputs, TestParseOutputsRejectsTruncatedDescriptor, TestIndicatorWriteKeepsOffHookBit, TestSetReportsUnadvertisedUsageWithoutWriting, TestFallbackOffHookWritesFullStandardReport, TestConcurrentWritersNeverEmitATornReport, FuzzParseOutputs seed corpus)
- CGO_ENABLED=0 go test ./internal/hid/... -fuzz FuzzParseOutputs -fuzztime 20s (1,247,626 execs, no crashes)
- git diff a69dcc7 72cdec6 -- internal/hid/telephony.go internal/hid/telephony_test.go (confirmed empty diff, F1/F2 untouched)
- Standalone scratch-module (outside repo, deleted after use) mutation check: reverted descriptor.go's new cap to the old fixed maxIndexedBits ceiling, re-ran TestUsageRangeCannotOutrunDescriptorLength -> FAILS as expected (16-byte descriptor indexes 32768 usages unrejected), confirming the fix is load-bearing
- Standalone scratch-module tests (written by reviewer, not in repo): TestCapAccumulatesAcrossOutputItemsSameReport and TestCapAllowsAccumulationUnderRatio, both PASS, verifying cross-item cap accumulation and legitimate multi-item report parsing
- CGO_ENABLED=1 go test -race ./... attempted, build failed: fatal error: /usr/include/stdc-predef.h: Permission denied (cgo blocked by sandbox, not executed)

jerryfane:
## Exact-head verdict: approved at `72cdec6`

I was wrong to describe this PR as approved after pushing `72cdec6`. The approval was at `a69dcc7`, and answering a P3 by pushing a commit moved the head past the verdict — the same superseded-head problem I had correctly cancelled an earlier job over. The coordinator caught it (directive 173576). This is the missing coverage, at the actual head.

| Field | Value |
| --- | --- |
| Verdict | `approved` |
| Reviewer | `voice-review-claude`, review-only (`--no-fix-target`, `--org-role voice`) |
| Job | `local-review-review-router-18d6297d58c6b2d6-1` |
| Head reviewed | `72cdec61647927f5250e6cc8197b189e95f3512b` — the current PR head |
| Scope | The delta commit against the previously reviewed `a69dcc7`, touching only `internal/hid/descriptor.go` and `descriptor_test.go`, plus re-confirmation that F1, F2 and F3 are unchanged and still correct |
| Prior findings | `#8-f1` … `#8-f4`, all `answered` |
| Model family | Same family as the implementation: codex returns `401 Unauthorized`, kimi has no credential on this host. |

`#8-f4` is resolved: the fixed 32768-bit ceiling is replaced by a length-proportional cap of eight indexed single-bit outputs per descriptor byte, and the comment now states that bound instead of the false "one usage per two bytes" claim.

### Review-head chain, for the record

1. `local-review-review-router-18d62657f49a5bae-1` at `87a938c` — **cancelled** by me, no recorded result, carries nothing forward.
2. `local-review-review-router-18d6269f619d09f4-1` at `2625ed3` — `changes_requested`, three findings.
3. `local-review-review-router-18d627cc59b69d43-1` at `a69dcc7` — `approved`, one new P3 (`f4`).
4. `local-review-review-router-18d6297d58c6b2d6-1` at `72cdec6` — **`approved`, this verdict, the current head.**

PR #11 is stacked on this branch, so if this head moves again, #11's review head must be revisited too.

Merge is not requested and I hold no merge authority here. The coordinator queues that decision for the owner, with merge order #8 first, then #11 rebased onto `main` and re-checked. The on-device criterion of #2 remains blocked on the owner, not failed.

PR #11, opened 2026-09-17
Run timers locally, with no model in the path

jerryfane:
Partial for #3 (deliberately not "closes"); supersedes my duplicate issue #10, now closed. Base is `main`: #8 was squash-merged as `8a11b49` and this branch was rebased onto it.

## What changed

The timer vocabulary is parsed, scheduled and announced by the session layer. No model, no network, no agent session:

- `internal/timer` — `Scheduler` (wall-clock deadlines, sleeps until the earliest one, wakes on change), `Store` (atomic JSON persistence), and spoken-duration parsing.
- `internal/session/timers.go` — set, query, cancel, expiry announcements, and the report for timers that came due while Voice was off.
- `timers.{enabled,file}` config; `voice doctor` reports how many are running and how many expired offline.
- `packaging/voice.service` gains `StateDirectory=voice` and `XDG_STATE_HOME=/var/lib`, so state is `/var/lib/voice/timers.json` owned by `voice-agent`.

Anything the timer vocabulary does not recognise reaches the planner unchanged. "Set a timer" with no duration asks how long rather than inventing a length.

## Why the scheduler looks like this

- **Wall-clock deadlines, not countdowns.** A suspended or slow host must not shift when a timer fires.
- **Atomic rename on every change.** A truncated write during a power cut would lose every timer.
- **Expired-while-down is reported, not fired.** Announcing a timer that finished twenty minutes ago as if it just finished is a lie; the report says how long ago, once, and the state is persisted before announcing so a restart loop cannot repeat it.
- **Independent of how a job was created**, because alarms and reminders in #1 need the same machinery.

## Verification

Session-layer and fake-capture only, per the coordinator's constraint: nothing here claims on-device behaviour, and no claim depends on #9.

With the built binary, a paced fake microphone (20 ms frames, like a real sound card), a fake STT that returns one sentence, and a brain command that logs if it is ever called:

| Scenario | Result |
| --- | --- |
| "hey voice set a timer for 3 seconds" | `timer set id=0 duration=3s`, said "3 seconds timer started.", fired at the deadline, said "Your 3 seconds timer is done.", state file emptied. **Brain called 0 times.** |
| 6 s timer, `SIGKILL` at 3 s, restart with a silent mic | Set 18:33:45, deadline 18:33:51, killed, restarted — fired at 18:33:51, the original wall-clock time |
[...]

PR #14, opened 2026-09-17
Leave no blank-discarded call result in production

### LINKED ISSUE #12: Invariant: no blank-discarded call result in production
Filed on its own rather than added to #11, which has had five review rounds already.

## Why an invariant rather than another rule

PR #11's review found the same defect class five times, and each round the fix was an enumeration:

1. fail on blank assignments taking a call result — missed **bare call statements** and **short declarations into a blank**, and four real `Close` discards were sitting in the repository while the check reported clean;
2. add bare calls and trailing blanks for a **named set** of operations — missed **`defer` and `go`**, which is the shape most `Close` calls actually take, and hid a real defect: the hidraw write dropped its close error, the one that keeps the microphone un-gated;
3. add `defer`/`go` — the **name list** then missed three `json.Marshal` discards, one of which would have handed the planner an empty device inventory and let it confidently deny that a device exists.

Every round enumerated something and the next real defect arrived in the gap. The invariant that ends the pattern does not enumerate: **production contains no blank-discarded call result**, checkable with `errcheck -blank -ignoretests` and **no exclude file**.

A type-level exclude file was considered and rejected: entries like `(*os.File).Close` silence the next real case at that type anywhere in the repository, which is the same false comfort in a new place.

## What this needs

Branch `no-blank-discards` has it, verified clean:

- `errcheck -blank -ignoretests ./...` in CI with nothing excluded;
- temp-file cleanup in the speech engines reports through a `faults.Reporter`, now required by their constructors;
- `rows.Close` and `stmt.Close` join the returned error (`Load`/`Save` take named results);
- `tx.Rollback` becomes an explicit check filtering `sql.ErrTxDone` by name, since that is the expected outcome after a successful commit and anything else means a transaction was left behind;
- the hidraw descriptor-size handle and the Magic Home connection join their close errors into the answers they return;
- the WAV encoder checks its writes and panics on the impossible, stating the `bytes.Buffer` judgement at the call;
- the in-suite AST check stays for the test tree CI skips, with its comment corrected to say it is a fast name-scoped guard rather than a complete one.

## Read `9096619` correctly

The commit in #11 that made `errcheck` default mode pass rewrote eight `defer x.Close()` into `defer func() { _ = x.Close() }()`. **That is a linter-clean intermediate, not a fix.** It moves the discard to where the linter cannot see it while the failure is just as lost. It is accepted as an intermediate because it is reviewed, CI-green and makes nothing worse — but nobody should later read that commit as the resolution of this class. This issue is the resolution.

## Acceptance criteria

- [ ] `errcheck -blank -ignoretests ./...` passes with no exclude file and no inline suppressions.
- [ ] No `_ =` of a call result remains in non-test code.
- [ ] Deferred cleanup failures reach a `faults.Reporter` wherever one is in scope, and are joined into the returned error where the function has one.
- [ ] `go.mod` and `go.sum` still contain no reference to the linter; the static `CGO_ENABLED=0` `arm64` build is unaffected.
- [ ] The in-suite check still passes and still says what it is scoped to.

jerryfane:
Closes #12.

## The invariant

Production contains **no blank-discarded call result**. CI enforces it with `errcheck -blank -ignoretests ./...` and **no exclude file**.

That replaces four rounds of enumeration on PR #11, each of which was answered with a longer list and each of which leaked:

| Round | Rule added | What arrived in the gap |
| --- | --- | --- |
| 1 | fail on blank assignments taking a call result | bare call statements and short declarations — **four live `Close` discards** the guard reported clean over |
| 2 | add bare calls and trailing blanks, for a named set | `defer` and `go`, the shape most `Close` calls take — hid the **hidraw write dropping its close error** |
| 3 | add `defer`/`go` | the name list missed **three `json.Marshal` discards**, one handing the planner an empty device inventory so it could deny a device exists |

An exclude file was considered and rejected: type-level entries like `(*os.File).Close` silence the next real case at that type, which is the same false comfort relocated.

## The conversion

- Temp-file cleanup in the speech engines reports through a `faults.Reporter`, now required by their constructors — a leftover file nobody mentions is the silent-failure class `faults` exists for.
- `rows.Close` and `stmt.Close` join the returned error; `Load` and `Save` take named results to do it.
- `tx.Rollback` becomes an explicit check filtering `sql.ErrTxDone` **by name**: that is the expected outcome after a successful commit, and anything else means a transaction was left behind.
- The hidraw descriptor-size handle and the Magic Home connection join their close errors into the answers they return.
- The WAV encoder checks its writes and panics on the impossible, stating the `bytes.Buffer` judgement at the call and turning a future change of sink into a loud failure rather than a silent truncation.

## The in-suite check, and its own failure modes

`internal/codecheck` stays, for the test tree CI skips, with its comment corrected: it is a fast **name-scoped** regression guard, not a complete one, and the errcheck run is what makes completeness a property of production rather than of a list.

Because that guard has been wrong three times about what it can see, its failure modes are now assertions rather than something to trust — directive 173659 required a parse error to fail rather than be skipped, and that held only as a throwaway mutant before:

- an unparseable file fails the check, and the error names the file;
- a tree with no Go files fails, because a check that examined nothing would pass over anything;
- an unjustified discard is reported; a justified one is not.

## How to read `9096619` in #11

That commit rewrote eight `defer x.Close()` into `defer func() { _ = x.Close() }()`. **Incomplete, not cosmetic** — and this is a correction to my own earlier wording as well as to the coordinator's. The reviewer's comparative run showed errcheck's *default* mode flagged those seven sites at the prior head, including the `tx.Rollback` the in-suite check never covered, so the conversion fixed real violations rather than merely satisfying a stricter mode nobody runs. What remains true is that the failure is still discarded there; this PR is the resolution.

## Verification

At head `a748488`, rebased onto `697f6bc`:

- `go build ./...` 0, `go vet ./...` 0, `gofmt -l .` empty;
- `go test ./...` 0 and `CGO_ENABLED=1 go test -race ./...` 0;
- `errcheck -blank -ignoretests ./...` **exit 0 with nothing excluded**;
- `CGO_ENABLED=0 GOOS=linux GOARCH=arm64` builds a statically linked binary, 10,807,276 bytes;
- `go.mod`/`go.sum` unchanged — the linter is invoked as `go run pkg@pinned` and never enters the module.

No on-device claim: this touches error handling, not audio behaviour.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d630586c6c3bee-1`

**Decision:** `approved`
**Summary:** PR #14 at a748488 replaces the enumerated blank-discard guard with `errcheck -blank -ignoretests` (no exclude file) plus a name-scoped in-suite test-tree check. Independently executed build/vet/fmt/test all pass; -race is blocked by a host-level sandbox restriction (matches prior PR #8 review). Verified via Go stdlib source tracing that the tx.Rollback/sql.ErrTxDone handling in timer/store.go is exactly correct and a genuine post-commit-failure rollback failure is structurally unreachable via database/sql, so nothing is masked. Verified no handle leaks in descriptorSize/roundTrip, and confirmed the WAV-encoder panic is unreachable under all current call sites (defensible MustCompile-style assertion, not wrong). Found two non-blocking, evidence-backed nuances: (1) errcheck's own bundled ~19-symbol default-exclude list still applies despite 'no exclude file', surfacing 17 conventionally-unchecked fmt.Print-family stdout/stderr calls plus 2 provably-safe strings.Builder writes when disabled via -excludeonly; (2) a pre-existing (not introduced by this PR) off-by-one in internal/codecheck's justifiedLines lets a trailing justification comment on one line silently also justify an unrelated, unjustified discard on the next line, proven via an isolated scratch-module probe of the PR's own discard_test.go. Neither reduces actual production protection since errcheck (the authoritative check) ignores comments and default-excludes only provably-safe or conventional-output symbols outside the fixed defect class this PR targets. Same-model-family caveat: reviewed by a Claude runtime; applies if the implementing seat was also Claude.

**Findings**
- **errcheck's bundled DefaultExcludedSymbols still apply with 'no exclude file'** (P3)
  - description: errcheck v1.9.0 always merges ~19 built-in excluded symbols (bytes.Buffer/strings.Builder writes, fmt.Print family, io.Pipe CloseWithError, math/rand.Read, hash.Write) unless -excludeonly is passed, regardless of whether an exclude file exists. Most of these are provably safe (Builder/Buffer writes can't fail), but the fmt.Print-family stdout/stderr CLI output calls in cmd/voice/main.go can genuinely fail (broken pipe) and are unconditionally unchecked. Non-blocking: none of these are in the state-loss/handle-leak/inventory defect class this PR targets, but the 'no exclude file = nothing hidden' framing should carry this caveat.
  - locator: cmd/voice/main.go:26,41,65,80,94,101,129,134,148,166,167,178,190,224,226,228,272; internal/timer/parse.go:316,318 (proven via `errcheck -blank -ignoretests -excludeonly -exclude <empty>` surfacing all 19 sites that the actual CI command silently accepts)
- **Trailing justification comment silently justifies the following unrelated discard** (P3)
  - description: justifiedLines marks both the comment's own line and the following line as justified, to let a comment written above a statement cover it. The same rule also lets a trailing same-line comment leak coverage onto the next, unrelated line. This is pre-existing, not introduced by this PR, and does not reduce actual production protection since the authoritative errcheck run ignores comments entirely and would still catch a truly unjustified production discard; it only weakens the secondary, already-documented-as-incomplete in-suite guard on the test tree.
  - locator: internal/codecheck/discard_test.go justifiedLines() (function body unchanged by this PR's diff, inherited from base 697f6bc); reproduced via an isolated scratch-module probe (not committed, reviewed repo left untouched) with two consecutive `_ = f()` discards where only the first carries a same-line `// discard:` comment — scan() reported 0 offences instead of 1

**Tests Run**
- gofmt -l . -> empty, exit 0
- CGO_ENABLED=0 go test ./... -> ok, all packages, exit 0
- CGO_ENABLED=0 go test ./internal/timer/... -v -run 'TestStore|TestSave|TestLoad' -> PASS (TestSaveReplacesTheStoredSetTransactionally, TestSaveFailureAfterFiringIsReported)
- CGO_ENABLED=1 go test -race ./... (sandbox explicitly disabled) -> build failed on all packages, 'fatal error: /usr/include/stdc-predef.h: Permission denied' in runtime/cgo; host-level restriction, not executable in this environment
- CGO_ENABLED=0 go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank ./... -> exit 0, no output (production clean)
- CGO_ENABLED=0 go run github.com/kisielk/errcheck@v1.9.0 -blank ./... (no -ignoretests) -> 6 discards reported, all in internal/device/magichome_test.go and internal/timer/timer_test.go (test-only, expected)
- CGO_ENABLED=0 go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly -exclude <empty file> ./... -> 19 additional production discards surfaced (cmd/voice/main.go fmt.Print family x17, internal/timer/parse.go strings.Builder.WriteRune x2), proving these are silently absorbed by errcheck's own bundled defaults under the actual CI invocation
- Isolated scratch-module probe: extracted internal/codecheck/discard_test.go byte-for-byte via `git show a748488:...` into a throwaway module outside the reviewed repo, added one probe test with two consecutive unjustified-looking `_ = f()` calls (first carrying a trailing justification comment), ran `go test -run TestProbeTrailingJustificationLeaksToNextLine -v` -> scan() reported 0 offences instead of 1, confirming the justification-leak gap; scratch module deleted, `git status` on reviewed repo clean before and after
- Manual source trace of go1.26.4 database/sql's Tx.Commit/Rollback (sql.go) confirming tx.done is set unconditionally before the driver-level commit runs, so a genuine post-commit-failure rollback failure is structurally unreachable, validating the sql.ErrTxDone filter in timer/store.go Save()
- CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o <scratch> ./cmd/voice -> succeeded, static ELF aarch64, 10549189 bytes (vs 10807276 bytes claimed; attributed to Go toolchain version drift, go1.26.4 here vs whatever version produced the original evidence, not a functional defect)

Issue #9, opened 2026-09-17
Diagnosis blocked by an ambiguous log: capture worked, the transcriber could not start (cause: #13)


jerryfane:
Found while gathering hardware evidence for #2. Filed separately because it is not caused by #2 and it blocks the on-device smoke test that #2's acceptance criteria require.

## What was observed

Evidence collected by the pi-burj seat on the installed Raspberry Pi (Anker PowerConf, USB `291a:3301`, `/dev/hidraw5`, `plughw:2,0` in and out, unit user `voice-agent`):

- `journalctl -u voice --since "-7 days" | grep -o "heard=.*peak=[0-9.]*"` → no output. The session loop logs `heard=... peak=...` for every segmented utterance, so seven days of journal contain no utterance at all.
- `arecord -D plughw:2,0 -f S16_LE -r 16000 -c 1 -d 3` with the service stopped and the telephony off-hook report set (`02 01 00`) → `peak_abs=0`, `rms=0.000000`.
- Same capture with off-hook plus the mute LED bit (`02 03 00`) → `peak_abs=0`.
- Same capture repeated under the unit's real credentials (`setpriv --reuid=995 --regid=989 --groups=29,44,989`) → `peak_abs=0`.
- Same capture while a generated 1 kHz tone played out of the same PowerConf via `aplay -D plughw:2,0` → `peak_abs=0`.

## What this does and does not establish

It does **not** establish that the microphone is broken: none of these captures has a positive control, so "the device is silent" and "our capture path never reached the device" are both consistent with the data. Exact `0` across every configuration, including with a tone playing, is more typical of a capture that never received samples than of a live but quiet microphone, which would still show a noise floor.

It does establish that **no evidence exists that production capture works**, and therefore that any LED or earcon smoke test on this device currently proves nothing.

## Candidate causes, none confirmed

1. Voice closes the hidraw descriptor immediately after writing the off-hook report (`internal/audio/command.go`, `setOffHook`). If the PowerConf firmware treats the host as on-hook again when the descriptor closes, capture would be gated in production. Test: hold the descriptor open for the duration of a capture and compare peaks.
2. Card index drift: `plughw:2,0` may not be the PowerConf when the service is stopped.
3. The measuring path itself (no positive control yet). Test: capture from any other input device and confirm a non-zero peak.
4. Firmware requires a different report or ordering than report 2 bit 0.

## Next steps

- [ ] Positive control: non-zero peak from some capture on that host.
[...]

PR #15, opened 2026-09-17
Ship a speech engine that can start, and a log that says which stage stopped

### LINKED ISSUE #13: whisper-cli cannot load libwhisper.so.1, so no utterance is ever transcribed
Found by the pi-burj seat while confirming staging for #9.

## Observed

Every segmented utterance fails to transcribe on the installed Pi:

```
voice: transcribe: whisper-cli: exit status 127: whisper-cli: error while loading shared libraries:
libwhisper.so.1: cannot open shared object file: No such file or directory
```

`systemctl is-active voice` is `active`, and these are the newest lines in the journal.

## Why this is separate from #9, and worse than it looks

`Transcribe` is only called on a segmented utterance, so these lines prove the capture path delivered audio. The failure is downstream: the recognition binary cannot load its own shared library, so **every** spoken command dies before the wake gate is ever evaluated.

It also masked #9's evidence. `heard=` is logged only after a successful transcription, so `grep -c "heard="` returned 0 for the whole boot and I read that as "no audio was ever captured". A broken STT binary and a dead microphone produce the identical log signature, which is a diagnosability defect in its own right: the session layer logs nothing when an utterance is segmented but fails to transcribe beyond the engine's own error.

## Candidates

1. Stale loader cache — `libwhisper.so.1` present but not on the loader path, fixed by `ldconfig`. Cheapest to check, and would mean the deployment step is missing it.
2. Library genuinely absent — `scripts/setup-speech.sh` builds whisper.cpp; if the shared library is not copied to where the restricted `voice-agent` account can load it, the installer is incomplete.
3. Unit environment — the service sets `PATH` but not `LD_LIBRARY_PATH`, so a binary that works in an interactive shell can fail under systemd.

Diagnosis dispatched (read-only): `ldd $(command -v whisper-cli)`, `ls -l /usr/local/lib/libwhisper* /usr/lib/libwhisper*`, `cat /etc/ld.so.conf.d/*.conf`, and `systemctl show -p Environment -p ExecStart voice`.

## Acceptance criteria

- [ ] Root cause identified as loader path, missing artifact, or unit environment — not guessed.
- [ ] `voice` transcribes a spoken utterance on the installed device, with a `heard=` line and a non-zero peak in the journal.
- [ ] Whichever component is at fault is fixed where it belongs: `scripts/setup-speech.sh`, `scripts/install-local.sh`, or `packaging/voice.service`.
- [ ] A segmented utterance that fails to transcribe logs enough to tell that case apart from silence, so this cannot mask a capture problem again.
- [ ] `voice doctor` reports the speech-to-text engine as unusable when its library cannot load, rather than OK on the strength of the binary existing.

jerryfane:
Closes #13.

**The installed device has never transcribed a single utterance.** `whisper-cli` cannot load `libwhisper.so.1`, so every segmented utterance died at exec time — once per utterance, for the whole life of the boot. This is the product not working, not a missing feature.

## Which layer, and why

The defect is in the **installing layer**, not the unit. `scripts/setup-speech.sh` built whisper.cpp with cmake defaults, which produce **shared** libraries, then installed only the binary — so it went looking for a library that was never put anywhere. Evidence from the device:

```
$ ldd "$(command -v whisper-cli)" | grep -i whisper
libwhisper.so.1 => not found
$ ls -l /usr/local/lib/libwhisper* /usr/lib/libwhisper*
ls: cannot access ...: No such file or directory        # and ldconfig -p has no entry
```

Exporting `LD_LIBRARY_PATH` in `packaging/voice.service` would have made the symptom disappear while leaving a binary that breaks for anyone running it by hand, and would have put the fix a layer away from the mistake. The build now passes `-DBUILD_SHARED_LIBS=OFF`, so there is one file to install and no loader configuration to get wrong.

## The installer could not fail

It verified itself with `whisper-cli --help | sed -n '1p'`. A pipeline reports the status of its **last** command, so `sed` succeeding meant the script reported success while shipping an engine that could not start. Each engine is now executed and its exit status checked, with the loader's own message called out by name and a non-zero exit failing the script.

## The expensive defect was the log

A dead microphone and a broken transcriber produced **identical** journals: the only line was `heard=`, emitted *after* a successful transcription. That ambiguity is what cost #9 five rounds of reasoning — I concluded the microphone had never captured anything when it was in fact being heard.

Each stage now reports its own outcome, so a future silence names the stage that stopped:

```
stage=capture state=starting device="command input \"arecord\" on plughw:2,0 (16000 Hz mono)"
stage=segment state=ok samples=35200 ms=2200 peak=0.5787 truncated=false
stage=transcribe state=failed err=exit status 127: libwhisper.so.1: cannot open shared object file
```

That is the real binary reproducing the device's exact failure. Before this change those three lines were one absent line.

## `voice doctor` called it OK

It checked that a file existed on `PATH`. It now runs the engine: a binary that exits 127, or whose loader complains, is reported unusable —

```
speech-to-text FAIL whisper.cpp: /tmp/binshim/whisper-broken: whisper-cli: error while
loading shared libraries: libwhisper.so.1: cannot open shared object file
```

— while an engine that merely disagrees about the exit status of `--help` is still accepted, since engines differ there. Same class as every guard fixed this week: a check that cannot fail looks exactly like a healthy system.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean; `go test ./...` and `CGO_ENABLED=1 go test -race ./...` green.
- `sh -n scripts/setup-speech.sh` clean.
- New tests: the runnability probe rejects a shim that reproduces the loader failure verbatim, accepts a working engine, and accepts one that exits 1 on `--help`; two session tests assert every stage is observable and that a transcribe failure is distinguishable from silence.
- Real binary against a shim reproducing the device's failure: the journal shows `segment state=ok` with a non-zero peak followed by `transcribe state=failed`, and `doctor` reports the engine unusable.

## Two things named rather than smoothed over

**The static-build step is reasoned, not executed by me.** I cannot build whisper.cpp in this environment, so `-DBUILD_SHARED_LIBS=OFF` is a conclusion from the failure mode (cmake defaults produce shared libraries; only the binary was installed; `libwhisper.so.1 => not found` with nothing in any loader path) rather than something I ran. pi-burj's rebuild on the device is the **first real execution of this change**, and its result decides whether this PR is correct. It has since rebuilt statically and the device now transcribes, which is corroboration — but the repo path itself still has not been executed end to end, and I am asking for the exact cmake line it used so the two can be compared.

**The unfailable installer check is the third artifact this week whose check could not fail.** The pattern, in order: `voice doctor` reported the speech engine OK because a file existed on `PATH`; the in-suite discard guard reported clean while four real `Close` discards sat in the tree, twice, because it could not see the shapes they were written in; and this installer reported success because it verified itself through a pipeline, so `sed`'s exit status masked an engine that could not start. Each one looked exactly like a healthy system. That is the class, not three coincidences, and it is why the fix here is *make the check capable of failing* rather than *fix the one instance*.

## Not verified here

The fix on the device itself. pi-burj is rebuilding whisper.cpp under the owner's authorisation; whether the rebuild plus this change produces a `heard=` line on real speech is the acceptance test, and it runs with the owner present. I have not claimed it.


jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d630b9f88a406c-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d630b9f88a406c-1`

**Summary:** PR #15 (head 86bcf4e) is reviewed against its three stated goals plus the operator's directive to attack the probe. Static build reasoning holds up against whisper.cpp v1.9.4's real CMakeLists (WHISPER_BUILD_EXAMPLES is independent of BUILD_SHARED_LIBS; no ggml backend defaults shared) though the actual compile could not be run here (no C toolchain in this sandbox) - matches the PR's own disclosure. The installer's set -eu/if-condition semantics correctly avoid re-introducing the 'pipeline can't fail' bug for the engine-can't-run case (verified live against a shim reproducing the loader failure: exit 1 propagates correctly), but status=$? after the if/fi with no else always captures 0, not the real exit code (cosmetic only, diagnostic text is always wrong). Stage-by-stage logging is verified to distinguish a broken transcriber from true silence and does not flood the journal or leak transcript content beyond pre-existing behavior. The doctor probe correctly avoids the stdin-hang class (Stdin defaults to /dev/null, verified empirically) and is only ever invoked from the one-shot doctor command, never in a loop, but its claimed 5s timeout does not actually bound wall-clock time: exec.CommandContext only kills the direct child, and I reproduced a 30s hang via a backgrounded grandchild that inherits the output pipe even though the direct child exits cleanly - a real gap against the PR's own claim, reachable via user-configured engine commands even though the two default-installed engines wouldn't trigger it today. Separately, the shell installer's own check_engine has no tolerance for a non-zero --help exit with real output, unlike the Go-side runnable() which explicitly designs and tests for that - an inconsistency with the PR's own stated rationale, though it doesn't bite with the currently pinned engine versions (whisper-cli confirmed to exit 0 on --help from source). Build/vet/gofmt/tests/errcheck/cross-compiled static binaries all pass under CGO_ENABLED=0, matching
[truncated]
**Severity:** `P2`

**Findings**
- **P2**
  - description: exec.CommandContext's 5s timeout only kills the direct child process; a descendant that inherits the stdout/stderr pipe keeps CombinedOutput() blocked past the timeout. Reproduced live: a launcher script that backgrounds `sleep 30 &` and exits cleanly (err=nil) still made runnable() block for the full 30s instead of ~5s. Only reachable via `voice doctor` (not the persistent service loop, confirmed by grepping all Available() call sites), and not triggered by the two engines this repo installs (whisper-cli, piper are plain binaries), but CommandTranscriber/CommandSynthesizer explicitly adapt 'any file-oriented STT executable' via a user-configured Command argv, so wrapper/launcher-style custom engines are a real, not hypothetical, trigger.
  - location: internal/speech/command.go:55-58 (runnable)
- **P3**
  - description: status=$? immediately after `if out=$(...); then ... fi` with no else always evaluates to 0 (the if-compound's own exit status per POSIX when neither branch body ran), not the failed command's real exit code. Verified empirically against dash (this system's /bin/sh) with /bin/false and a 127-exit loader-failure shim: printed message always says '(exit 0)' regardless of actual code. Diagnostic text only - the surrounding return 1/failed=1/exit 1 gate is unaffected and was verified to still correctly fail the installer.
  - location: scripts/setup-speech.sh:76 (status=$? inside check_engine)
- **P3**
  - description: check_engine treats any non-zero --help exit as fatal, with no tolerance for a non-zero exit accompanied by real (non-loader-error) output - unlike internal/speech/command.go's runnable(), which explicitly implements and tests exactly that tolerance per the PR's own stated rationale ('a non-zero exit with sensible output is not evidence of anything'). Verified live: a shim that prints real usage text and exits 1 is rejected by check_engine (failed=1) while the equivalent case is accepted by runnable() in Go. Confirmed whisper-cli exits 0 on --help from v1.9.4 source, so this doesn't currently misfire with the pinned engines, but it's an inconsistency with the PR's own design philosophy and a latent fragility if either pinned engine or version changes.
  - location: scripts/setup-speech.sh:70 (check_engine's if out=$("$path" --help 2>&1); then)

**Tests Run**
- cd <worktree> && CGO_ENABLED=0 go build ./... -> ok
- cd <worktree> && CGO_ENABLED=0 go vet ./... -> ok
- cd <worktree> && gofmt -l . -> clean, no output
- cd <worktree> && CGO_ENABLED=0 go test ./... -> ok, all 12 packages pass including internal/session and internal/speech
- cd <worktree> && CGO_ENABLED=1 go test -race ./internal/session/... ./internal/speech/... -> FAILED to build: cgo unavailable in this sandbox (fatal error: /usr/include/stdc-predef.h: Permission denied), reproduced even with a trivial gcc -c hello-world compile and with dangerouslyDisableSandbox=true; environment limitation, not a PR defect, and not part of this project's own CI either
- sh -n scripts/setup-speech.sh -> clean
- go run github.com/kisielk/errcheck@v1.9.0 -ignoretests ./... (this project's own CI step) -> exit 0, clean
- CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/voice && GOARCH=arm64 (this project's own CI 'Static binaries' step) -> both succeed
- Extracted check_engine() verbatim from scripts/setup-speech.sh and ran it under dash against: a shim exiting 127 with the literal loader message (returns 1, failed=1, script exits 1, confirming the gate itself is not broken); a working shim exiting 0 (accepted); a 'grumpy' shim exiting 1 with real usage output (rejected by check_engine, unlike Go's runnable(), demonstrating the inconsistency finding); confirmed status=$? always prints 0 regardless of actual exit code in all three cases
- Ephemeral Go test (written in a writable scratch copy of the tree, not the reviewed read-only worktree, and deleted afterward) exercising runnable() directly: against a hanging binary (sleep 30) -> correctly detected as unusable but took the full 30s wall-clock, not ~5s; against a stdin-reading binary (read line) -> returned usable in 2ms, confirming no stdin-hang risk since exec.Command defaults Stdin to /dev/null
- Minimal standalone Go repro isolating exec.CommandContext against a shell script that backgrounds sleep 30 & and exits cleanly itself -> CombinedOutput() blocked 30s despite ctx timeout of 3s and the direct child reporting err=nil, confirming the pipe-inheritance gap generalizes beyond the specific test shim
- grep -rn ".Available(" across all non-test .go files -> confirmed runnable()/Available() is only ever called from cmd/voice/main.go's doctor() command, never in the persistent session Run() loop or any looped context
- Fetched whisper.cpp v1.9.4's actual CMakeLists.txt and ggml/CMakeLists.txt from GitHub to verify WHISPER_BUILD_EXAMPLES is independent of BUILD_SHARED_LIBS and no ggml backend defaults to shared regardless of BUILD_SHARED_LIBS (GGML_BACKEND_DL requires BUILD_SHARED_LIBS=ON and defaults OFF) -> confirmed, static build reasoning holds; also fetched examples/cli/cli.cpp confirming whisper-cli calls exit(0) on --help
- Checked GITMOOT_PRIOR_VERDICTS file -> no prior verdicts recorded for PR #15; confirmed 'new ground, no prior findings' per requester instructions

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d631baf3431996-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/voice#15 at exact head 33e69be (diff since prior head 86bcf4e limited to internal/speech/command.go, internal/speech/runnable_test.go, scripts/setup-speech.sh). All three named findings (f1/f2/f3) were independently re-verified live and are fixed: the WaitDelay=1s fix collapses the reported 30s hang to ~1.01s and preserves correct classification and exit-code detection; the setup-speech.sh status capture now reports real exit codes under dash with set -eu; check_engine's tolerance rule now matches runnable()'s exactly across 7 shim scenarios I built and ran. One new, narrow P3 defect surfaced: WaitDelay unblocks the read but does not terminate a lingering descendant, so a launcher-style engine leaves an orphaned background process running past voice doctor's return (reproduced with ps aux). No P1/P2 remain open. Build/vet/tests executed and green under CGO_ENABLED=0; the race build and the GitHub CI check could not be executed in this sandbox (cgo/gcc header access and gh auth are both blocked here), so those two specific claims from the requester are unverified by me, not contradicted. Same-model-family caveat: this review, like the prior round being answered, was performed by a Claude model (Sonnet 5), so treat convergence between rounds with that in mind. Per the PR body's own caveat, the static whisper.cpp build step remains reasoned, not executed, by any reviewer to date; pi-burj's device rebuild is still its first real execution.
**Severity:** `P3`

**Findings**
- **Probe timeout fix (cmd.WaitDelay=1s) verified live; hang reduced from 30s to ~1s with correct classification preserved** (P3)
  - continues_uid: jerryfane/voice#15-f1
  - detail: Ran the checked-in regression test: `go test ./internal/speech/... -v -run Runnable` — TestRunnableReturnsOnTimeEvenIfTheEngineSpawnsALingeringChild passes in 1.00s (was 30s pre-fix). Independently re-implemented the same exec.CommandContext+WaitDelay pattern in a standalone Go probe outside the repo and ran it against a launcher shim backgrounding `sleep 33 &`: elapsed=1.003s, err="exec: WaitDelay expired before I/O complete", out captured correctly as "usage: launcher-engine\n". Confirmed runnable()'s tolerance rule (err!=nil but text!="" is not fatal) correctly classifies this as available=true — the right outcome for a real working engine. Also verified two follow-up questions the coordinator asked: (1) a legitimate slow single-process engine with no descendant (3s pause between two writes, no forking) returns the full, untruncated output at elapsed=3.01s with err=nil — WaitDelay's grace timer only starts once the direct child's own exit is observed or the context is done, so ordinary slow --help output is not cut off; (2) exit-code detection stays correct even combined with a lingering descendant — a launcher that backgrounds `sleep 10 &` then itself exits 127 still yields a matched *exec.ExitError with code 127 at elapsed=1.00s. WaitDelay (vs. a process-group kill) is the right mechanism here: it is the stdlib's documented answer to exactly this problem, and unlike a pgid kill it does not assume authority to terminate whatever the child intentionally spawned. Resolved as reported; residual gap filed as a new finding below.
  - file: internal/speech/command.go:60-63
  - state: answered
- **Diagnostic exit-status bug fixed: status now reflects the real failing exit code under dash with set -eu** (P3)
  - continues_uid: jerryfane/voice#15-f2
  - detail: Extracted check_engine() verbatim via sed and ran it under dash (this system's /bin/sh, matching the script's #!/usr/bin/env sh) against a shim that exits 3 with no output: printed exactly 'silent failed to run (exit 3) with no output', matching the requester's claim. Confirmed `out=$(...) || status=$?` under `set -eu` does not trip -e: wrapped the extracted function in a harness with `set -eu` at the top and it completed normally (no premature abort) across all five shim cases I ran (loader-fail/127, grumpy-usage exit1-with-output, clean exit0, silent exit3-no-output, plain127 command-not-found). The `||` correctly protects the assignment as part of an AND-OR list, so it cannot mask the surrounding return 1/failed=1/exit 1 gate, which still correctly fails the installer in every failing case tested. Fixed.
  - file: scripts/setup-speech.sh:76-77
  - state: answered
- **check_engine now applies the same tolerance rule as runnable(); verified equivalent across 7 shim scenarios** (P3)
  - continues_uid: jerryfane/voice#15-f3
  - detail: Ran the extracted check_engine() under dash against: loader-failure (rc=127, loader message) -> fatal; grumpy exit-1-with-real-usage-output -> now SUCCEEDS (previously rejected, this was the bug); clean exit-0 -> succeeds; silent exit-3-no-output -> fatal; plain exit-127-no-output -> fatal; loader message appearing on line 2 of multiline output -> still correctly detected as fatal; nonexistent path (exit 127 via shell 'not found') -> fatal. Every outcome matches the rule implemented in internal/speech/command.go's runnable(): only a loader message, a 127, or a non-zero exit with NO output is fatal; a non-zero exit with real output is tolerated. The two checks no longer disagree about a working engine. Fixed.
  - file: scripts/setup-speech.sh:83-99
  - state: answered
- **WaitDelay unblocks the probe's pipes but does not terminate the orphaned descendant that was holding them** (P3)
  - detail: New observation, not previously named. Reproduced: after the launcher-engine probe (backgrounding `sleep 33 &`) returns in ~1s, `ps aux` still showed the sleep process alive and running to completion, now reparented as an orphan rather than being reaped or killed. Before this fix, the calling Go process blocked until the descendant naturally finished (undesirable, but the descendant was never left running unsupervised past the caller's return); after the fix, `voice doctor` returns promptly but leaves the background process running independently, once per invocation against such a wrapper/launcher-style custom engine. This does not affect the correctness of the boolean result (the tolerant text!="" check still correctly reports the engine as available), and the reachable caller remains only `voice doctor` (not the per-utterance service loop, unaffected by this narrow case), so I've scored it P3 rather than blocking. This directly answers the coordinator's question about WaitDelay vs. a process-group kill: a pgid-based kill would close this specific gap, but it introduces its own risk of terminating processes a launcher legitimately intends to keep running after --help returns, which WaitDelay deliberately avoids asserting authority over. Worth a comment or follow-up if wrapper-style custom engines become a supported/expected case rather than an edge case.
  - file: internal/speech/command.go:60-63

**Tests Run**
- cd workspace && CGO_ENABLED=0 go build ./... — success
- cd workspace && CGO_ENABLED=0 go vet ./... — clean
- cd workspace && gofmt -l . — no output, clean
- cd workspace && CGO_ENABLED=0 go test ./... — all packages pass, incl. internal/speech in 1.028s
- cd workspace && CGO_ENABLED=0 go test ./internal/speech/... -v -run Runnable — both tests PASS, lingering-child test elapsed 1.00s
- CGO_ENABLED=1 go test ./internal/speech/... -race -run Runnable — build failed in this sandbox: cgo requires gcc access to /usr/include/stdc-predef.h, which this sandbox denies (confirmed independently with a bare `gcc t.c -o t` also denied); race variant not executable here, not verified by me
- sh -n scripts/setup-speech.sh (dash) — clean, no syntax errors
- Standalone dash harness extracting check_engine() verbatim from scripts/setup-speech.sh and running it under set -eu against 7 constructed shims (loader-fail rc127, grumpy-usage rc1-with-output, clean rc0, silent-fail rc3-no-output, plain127, multiline loader message, nonexistent path) — all outcomes matched expected fixed behavior and runnable()'s tolerance rule
- Standalone Go probe (outside repo tree) replicating exec.CommandContext + cmd.WaitDelay=1s against: (a) launcher backgrounding sleep 33s+echo+exit0 — elapsed 1.003s, output captured, correctly tolerant; (b) slow single-process engine with a 3s mid-output pause, no descendant — elapsed 3.01s, full untruncated output, err=nil, no cutoff; (c) launcher backgrounding sleep 10s then exiting 127 itself — elapsed 1.00s, ExitError code 127 correctly detected
- ps aux after (a) above — confirmed the backgrounded sleep process was left running as a live orphan, not killed by WaitDelay
- git status --short / git log -1 on the workspace — confirmed exact head 33e69be9268c348cb95f57ef93289e8a89fc9547 and a clean tree (no edits made by this review)

jerryfane:
## Approved at the current head `33e69be9268c348cb95f57ef93289e8a89fc9547`

| Field | Value |
| --- | --- |
| Verdict | `approved` |
| Reviewer | `voice-review-claude`, review-only (`--no-fix-target`, `--org-role voice`) |
| Job | `local-review-review-router-18d631baf3431996-1` |
| Head reviewed | `33e69be9268c348cb95f57ef93289e8a89fc9547` — the current PR head, CI `test` SUCCESS |
| Prior round | `changes_requested` at `86bcf4e`, three findings, all re-verified live as fixed |
| New | `#15-f4` (P3) — **not fixed here**, disposition below |

Findings `f1`–`f3` are confirmed resolved: the probe's 30 s hang collapses to ~1 s with classification preserved, the installer's exit status now reflects the real failing code under `dash` with `set -eu`, and `check_engine` matches `runnable()` across seven shim scenarios.

### `#15-f4` (P3) — the orphaned descendant, deliberately deferred

Correct and newly observed: `WaitDelay` unblocks the probe's pipes but does not terminate the descendant that was holding them, so a launcher-style engine leaves a background process running unsupervised, once per `voice doctor` invocation. Before the fix the caller blocked instead — worse for responsiveness, but the descendant was never left running past the caller's return.

**Not fixed in this PR, on purpose.** The proper fix is a process group (`Setpgid` plus a group kill on timeout), which is a real change to how the probe spawns processes and deserves its own review rather than a seventh round on an approved head. This PR is the difference between a device that has never transcribed anything and one that works; the orphan is a leaked `sleep` on a custom wrapper engine that no shipped configuration uses. Filed as its own issue.

The judgement being applied: churning an approved head has cost real review rounds this week, and the coordinator moved the blank-discard conversion out of #11 for exactly this reason. Deferring a P3 with a written disposition is the version of that lesson I can apply myself.

### Verification at this head

`go build`, `go vet`, `gofmt -l` clean; `go test ./...` and `CGO_ENABLED=1 go test -race ./...` green; `sh -n` on the installer clean; CI `test` SUCCESS. `check_engine` exercised against four shims (loader failure, non-zero exit with usage text, clean exit, silent exit 3) with the expected classification each time.

Still named rather than smoothed over: the static-build step is **reasoned, not executed by me**, and pi-burj's device rebuild is its first real execution — it has since rebuilt statically and the engine now starts, which is corroboration, not proof of the repo path. After merge, the device should be reinstalled *through* `setup-speech.sh` and its `cmake` line compared against what the script produces; if they differ, the script is not yet the fix.

jerryfane:
## Rebased onto `0dcf47e`; conflict resolved by keeping both test sets

New head `13c2a7be06bbf2539253609bc1d779b57ecee440`. The approved head `33e69be` no longer exists on this branch — the rebase rewrote it, so **that approval does not transfer** and this head needs its own verdict.

### The conflict

One hunk, `internal/session/session_test.go`, at the end of the file where both branches appended tests:

| Side | Test | Claim |
|---|---|---|
| `main` (#16) | `TestBareWakePhraseAsksAndSkipsThePlanner` | a standalone wake phrase is acknowledged and skips the planner |
| this branch (#15) | `TestEachPipelineStageIsObservable` | every stage logs which stage it was |
| this branch (#15) | `TestTranscribeFailureIsDistinguishableFromSilence` | a failed transcribe does not look like silence |

These are independent claims, so **both sides were kept**; the file went from 13 tests to 14 and no assertion was dropped. Taking either side wholesale would have silently deleted coverage that had already been reviewed and approved, which is the specific risk in resolving a test-file conflict by hand.

### Evidence at the new head

All five tests from both branches execute and pass by name:

```
--- PASS: TestBareWakePhraseAsksAndSkipsThePlanner (0.00s)
--- PASS: TestEachPipelineStageIsObservable (0.00s)
--- PASS: TestTranscribeFailureIsDistinguishableFromSilence (0.00s)
--- PASS: TestRunnableRejectsABinaryThatCannotStart (0.01s)
--- PASS: TestRunnableReturnsOnTimeEvenIfTheEngineSpawnsALingeringChild (1.00s)
```

`go build ./...`, `go vet ./...`, `gofmt -l` clean; `go test ./...` and `CGO_ENABLED=1 go test -race ./...` green; `sh -n scripts/setup-speech.sh` ok.

One note on method, since it is the theme of this whole PR. My *first* coverage check was `go test ./internal/... -run ^TestName$` per test, counting `ok` lines — which reports 10 packages `ok` whether or not the test exists, because a package with no matching test still passes. That is another unfailable check, written by me, while checking for lost coverage. The `-v` named-`PASS` output above is the one that can fail, and I confirmed it: a nonexistent test name yields zero `PASS` lines.

### Unchanged from the approved head

The three earlier findings and the `#15-f4` deferral read the same: `-DBUILD_SHARED_LIBS=OFF` in the installer, the installer failing instead of reporting success, per-stage `stage=` logging, and the `runnable()` probe with `cmd.WaitDelay`. The orphaned-descendant fix remains deferred to **issue #18**, where the acceptance criterion is that the test asserts the descendant is *gone*, not that the probe returned promptly.

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d632e2115dcd68-1`

**Decision:** `failed`
**Summary:** Error: piped stdin input exceeds 10MB. Pass large content as a file path in your prompt instead.: exit status 1
**Diagnostics:** Error: piped stdin input exceeds 10MB. Pass large content as a file path in your prompt instead.: exit status 1

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d6331cd54ea13a-1`

**Decision:** `failed`
**Summary:** Error: piped stdin input exceeds 10MB. Pass large content as a file path in your prompt instead.: exit status 1
**Diagnostics:** Error: piped stdin input exceeds 10MB. Pass large content as a file path in your prompt instead.: exit status 1

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d63335262fb8ae-1`

**Decision:** `failed`
**Summary:** Error: piped stdin input exceeds 10MB. Pass large content as a file path in your prompt instead.: exit status 1
**Diagnostics:** Error: piped stdin input exceeds 10MB. Pass large content as a file path in your prompt instead.: exit status 1

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d636c6ff2ec36c-1`

**Decision:** `approved`
**Summary:** Independently reviewed jerryfane/voice PR #15 at exact head 13c2a7be06bbf2539253609bc1d779b57ecee440 (confirmed via `git ls-remote` matching refs/pull/15/head and refs/heads/issue-13-speech-install). Could not read /root/voice-pr15-context.md (permission denied even as root/with sandbox disabled - genuine LSM restriction, not a real ACL issue per getfacl); verified everything from GitHub source instead, which corroborates the requester's narrative. Priority One (rebase conflict in internal/session/session_test.go): fetched real parents 0dcf47e and 33e69be9268c348cb95f57ef93289e8a89fc9547 into a scratch clone (workspace .git/objects is write-protected, correctly enforcing review-only) and diffed session_test.go at HEAD against both - both diffs are 100% additive with zero deletions, and the function-name inventory at HEAD is the exact union of all tests from both parents (main's TestBareWakePhraseAsksAndSkipsThePlanner plus this branch's TestEachPipelineStageIsObservable and TestTranscribeFailureIsDistinguishableFromSilence). No assertion, helper, fixture, or table row was lost in the hand resolution. Priority Two: all four earlier findings read the same at this head and were re-verified, several live - BUILD_SHARED_LIBS=OFF (scripts/setup-speech.sh:36), installer exit 1 on failed engines (:106-109), per-stage stage= logging (internal/session/session.go), runnable()'s WaitDelay=1s (internal/speech/command.go:77, confirmed by test executing in 1.00s not 30s). check_engine was run live against 5 shims (loader failure, exit 127, exit 1 with output, silent nonzero, clean exit) and matches runnable()'s classification exactly. The orphaned-descendant issue (#18) was independently reproduced with my own throwaway program: the probe returns in ~1.00s but the backgrounded child survives past return, only exiting on its own timer - confirming the deferral is real and matches its stated acceptance criterion (test must assert the descendant is gone, not just that the probe retur
[truncated]

**Findings**
- Finding
  - description: Rebase conflict resolution in internal/session/session_test.go verified correct: diffs against both real parents (0dcf47e and 33e69be) are purely additive with zero deletions; HEAD contains the exact union of all test functions from both branches. No coverage lost.
  - locator: internal/session/session_test.go (diffed against 0dcf47e7799e745453dd0664f499036d920312d0 and 33e69be9268c348cb95f57ef93289e8a89fc9547)
- Finding
  - description: f1-f3 confirmed present and functioning at this head: -DBUILD_SHARED_LIBS=OFF, installer exit-1-on-failure, check_engine/runnable() tolerance parity - reconfirmed live via 5 shell shim scenarios matching Go's classification exactly.
  - locator: scripts/setup-speech.sh:36,67-100,102-109; internal/speech/command.go:42-56,153-167
- Finding
  - description: f4 (orphaned descendant, issue #18) genuinely still open and correctly deferred, not silently fixed or silently broken further: independently reproduced that runnable()'s WaitDelay makes the probe return in ~1s, but a backgrounded descendant survives past that return until its own timer expires. Matches issue #18's acceptance criterion exactly (test must assert the descendant is gone, which it does not yet).
  - locator: internal/speech/command.go:67-93; internal/speech/runnable_test.go:54-70 (TestRunnableReturnsOnTimeEvenIfTheEngineSpawnsALingeringChild only asserts elapsed<=10s, not descendant death); GitHub issue #18
- Finding
  - description: Methodology note (Priority Three): beyond the already-fixed 'go test -run <name>' issue, whole-suite 'go test ./...' without -v/name matching has the same present-vs-absent blindness (reports ok per package regardless of how many tests exist in it), and 'sh -n scripts/setup-speech.sh' is syntax-only and would pass even if check_engine's body were gutted. Neither is a defect in the PR's code, but both are unfailable-by-omission checks that should not be cited alone as coverage evidence.
  - locator: scripts/setup-speech.sh (sh -n scope); go test ./... (suite-level, no -run/-v)
- Finding
  - description: CGO_ENABLED=1 and go test -race could not be executed in this sandbox: even a trivial C file fails to compile because /usr/include is unreadable (verified independently, including with sandbox override). This matches every prior reviewer of this PR and is an environment limit, not a PR defect. GitHub's CI check for this exact head independently reports 'test completed success'.
  - locator: environment: /usr/include/stdc-predef.h permission denied under gcc; GitHub check-runs API for commit 13c2a7be06bbf2539253609bc1d779b57ecee440

**Tests Run**
- CGO_ENABLED=0 go build ./... -> BUILD_OK
- CGO_ENABLED=0 go vet ./... -> VET_OK
- gofmt -l . -> clean (no output)
- CGO_ENABLED=0 go test ./... -> all packages ok (audio, codecheck, device, feedback, hid, session 2.793s, speech 1.028s, timer, vad, wake)
- CGO_ENABLED=0 go test ./internal/session/... -v -run 'TestBareWakePhraseAsksAndSkipsThePlanner|TestEachPipelineStageIsObservable|TestTranscribeFailureIsDistinguishableFromSilence' -> 3/3 PASS by name
- CGO_ENABLED=0 go test ./internal/session/... -v -run 'ThisTestDoesNotExist' -> 'no tests to run', zero PASS lines (confirms named-PASS check distinguishes present/absent)
- CGO_ENABLED=0 go test ./internal/session/... -run 'ThisTestDoesNotExist' (no -v) -> 'ok ... [no tests to run]', exit 0 (reproduces the flawed unfailable check)
- CGO_ENABLED=0 go test ./internal/speech/... -v -run 'TestRunnableRejectsABinaryThatCannotStart|TestRunnableReturnsOnTimeEvenIfTheEngineSpawnsALingeringChild' -> both PASS, second completes in 1.00s (confirms WaitDelay fix)
- git diff (scratch clone) 0dcf47e7799e745453dd0664f499036d920312d0..13c2a7be06bbf2539253609bc1d779b57ecee440 -- internal/session/session_test.go -> pure addition, 0 deletions
- git diff (scratch clone) 33e69be9268c348cb95f57ef93289e8a89fc9547..13c2a7be06bbf2539253609bc1d779b57ecee440 -- internal/session/session_test.go -> pure addition, 0 deletions
- sh -n scripts/setup-speech.sh -> syntax OK
- Live shell harness: extracted check_engine() and ran against 5 shims (loader-failure, exit-127, exit-1-with-output, silent-exit-3, clean-exit) -> classification matches Go runnable() exactly (fail/fail/pass/fail/pass)
- Custom Go repro (outside PR tree) of exec.CommandContext+WaitDelay+backgrounded child -> probe returns in 1.001s, but 'sleep 8' descendant process confirmed alive via ps immediately after and only exits ~7s later on its own timer, never killed by the probe (confirms issue #18 is real and unfixed here)
- gcc trivial #include<stdio.h> compile attempt -> fails with '/usr/include/stdc-predef.h: Permission denied', confirming CGO_ENABLED=1/-race is unexecutable in this sandbox
- curl GitHub API repos/jerryfane/voice/commits/13c2a7be.../check-runs -> 'test completed success'
- curl GitHub API repos/jerryfane/voice/pulls/15 -> mergeable_state=clean, base=0dcf47e, 2 commits
- curl GitHub API repos/jerryfane/voice/issues/18 -> confirms deferral text and acceptance criteria match exactly

PR #20, opened 2026-09-17
docs(setup): state that the whisper build is not static, only free of libwhisper

jerryfane:
## The installer's own comment overstated what the build produces

`scripts/setup-speech.sh` said:

> Linking statically means there is one file to install.

The first half is wrong, and the loose version of this claim is how the original defect reached the device in the first place — "it's static" is exactly the kind of sentence that stops anyone checking what the loader will actually look for.

### Evidence from the installed host

After pi-burj reinstalled through the repo path (merged `main`), `ldd /usr/local/bin/whisper-cli` on the arm64 device resolves:

```
linux-vdso.so.1
libgomp.so.1    => /lib/aarch64-linux-gnu/libgomp.so.1
libstdc++.so.6  => /lib/aarch64-linux-gnu/libstdc++.so.6
libm.so.6       => /lib/aarch64-linux-gnu/libm.so.6
libgcc_s.so.1   => /lib/aarch64-linux-gnu/libgcc_s.so.1
libc.so.6       => /lib/aarch64-linux-gnu/libc.so.6
```

The binary is dynamically linked. What `-DBUILD_SHARED_LIBS=OFF` actually changes is narrower, and is the part that matters: **whisper.cpp's own libraries** are linked into the executable, so there is no `libwhisper.so.1` for the loader to fail to find. Those distribution libraries ship with the base system, which is why the reported failure cannot recur.

### What changed

Only the comment — the build flags are unchanged and correct. It now states the distinction explicitly, so nobody reads "static" and concludes the binary carries no runtime dependencies.

`README.md` needed no change: its `static` references are the `CGO_ENABLED=0` Go builds, which genuinely are static, and it makes no claim about the whisper build at all. I checked rather than assuming.

### Why this is its own PR

#13 is closed and correctly so — pi-burj's hand `cmake` line and the script's line match on every flag, including `-DBUILD_SHARED_LIBS=OFF`, differing only in source and build paths. So the installer *is* the fix and the device is now installed through the repo path. This is a claim-accuracy defect in a comment, not a reopening.

`errcheck -ignoretests -blank -excludeonly` (exclusion-free, on merged `main`) exits 0; `sh -n scripts/setup-speech.sh` ok.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d639105f4ec6b4-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/voice PR #20 at exact head 568d7f913272d5820cb7d7f4129a3c0d88a1bd0b (single commit on top of e3b60a7, touching only scripts/setup-speech.sh: 10 insertions/1 deletion, every changed line a '#' comment). The new comment's technical claim was independently verified against whisper.cpp v1.9.4's actual CMake sources (not taken on the PR author's word): native (non-Emscripten/non-MinGW) builds default BUILD_SHARED_LIBS to ON, so the flag is genuinely load-bearing; src/CMakeLists.txt names the target 'whisper' with SOVERSION=WHISPER_VERSION_MAJOR(1), so an unflagged shared build would produce exactly libwhisper.so.1, matching the original defect. Requester's specific worry — that ggml might be a separate target immune to the flag, making the new comment incomplete in the same way the old one was wrong — does not hold: ggml/CMakeLists.txt's own option(BUILD_SHARED_LIBS ...) is a no-op once the value is already cached from the parent's -D flag (standard CMake option() semantics), ggml/src/CMakeLists.txt's add_library(ggml-base/ggml/backend) calls carry no explicit SHARED/STATIC keyword so they inherit the same global variable, and the one path that could force a backend to MODULE regardless (GGML_BACKEND_DL) defaults OFF and throws a CMake FATAL_ERROR if set ON while BUILD_SHARED_LIBS is OFF — so it cannot silently produce a shared ggml artifact under this script's exact flags. GGML_STATIC also defaults OFF and is not set by the script, which is consistent with (not contradicted by) the reported ldd output showing libstdc++/libgcc_s still resolved from the distribution. README's 'static' references were independently re-read at README.md:170-171 and confirmed to be exclusively about the CGO_ENABLED=0 modernc.org/sqlite Go builds; a separate grep for 'whisper' in README (lines 6, 26, 108) found no static-related claim there either, and .github/workflows/ci.yml's own 'Static binaries' step only builds ./cmd/voice with CGO_ENABLED=0, confirming that's the onl
[truncated]

**Tests Run**
- git show 568d7f913272d5820cb7d7f4129a3c0d88a1bd0b -- scripts/setup-speech.sh: confirmed diff is comment-only (11 '+' / 2 '-' diff lines including headers = 10 insertions/1 deletion per commit stat), cmake invocation line unchanged
- sh -n scripts/setup-speech.sh: exit 0, syntax valid
- curl -s https://raw.githubusercontent.com/ggml-org/whisper.cpp/v1.9.4/CMakeLists.txt | grep -n BUILD_SHARED_LIBS: confirmed BUILD_SHARED_LIBS_DEFAULT=ON for native Linux (non-Emscripten/non-MinGW) builds, i.e. shared is whisper.cpp's true default absent the flag
- curl -s https://raw.githubusercontent.com/ggml-org/whisper.cpp/v1.9.4/src/CMakeLists.txt | grep -n 'add_library|SOVERSION': confirmed target 'whisper' with SOVERSION=WHISPER_VERSION_MAJOR(=1), i.e. an unflagged shared build produces exactly libwhisper.so.1
- curl -s https://raw.githubusercontent.com/ggml-org/whisper.cpp/v1.9.4/ggml/CMakeLists.txt and ggml/src/CMakeLists.txt: confirmed ggml's add_library calls for ggml-base/ggml/backends carry no explicit SHARED/STATIC keyword and inherit the same top-level BUILD_SHARED_LIBS cache variable; GGML_BACKEND_DL (the only bypass) defaults OFF and requires BUILD_SHARED_LIBS=ON or CMake raises FATAL_ERROR
- grep -n -i static README.md and read lines 155-185: confirmed the only static references are the CGO_ENABLED=0 modernc.org/sqlite Go builds
- grep -n -i whisper README.md (lines 6, 26, 108): confirmed no static-related claim about the whisper build anywhere in README
- cat .github/workflows/ci.yml: confirmed the 'Static binaries' CI step builds only ./cmd/voice with CGO_ENABLED=0, unrelated to whisper.cpp
- CGO_ENABLED=0 go build ./...: exit 0
- CGO_ENABLED=0 go vet ./...: exit 0
- CGO_ENABLED=0 go test ./...: all 8 tested packages ok, exit 0
- test -z "$(gofmt -l .)": clean, no output
- go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... (exact CI-pinned version): exit 0, no findings
- CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /dev/null ./cmd/voice: exit 0
- CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o /dev/null ./cmd/voice: exit 0
- python3 scan of /root/.gitmoot/cache/tools/read-only/ba5f5ea6956edf15/evidence/prior-verdicts.json: no entries for pull_request 13 or 20 among all 20 recorded verdicts

PR #25, opened 2026-09-18
fix(install): restart the service and verify the running program is the installed one

### LINKED ISSUE #24: install-local.sh reports success while the old binary keeps running (enable --now does not restart)
## Observed

The device has been running a `voice` binary built before PR #15: no `stage=` lines appear in the journal at all, while the checkout contains them. Reinstalling through `scripts/install-local.sh` did not change that — after the install, the running PID kept emitting the old log format.

## Cause

`scripts/install-local.sh` ends with:

```sh
sudo systemctl enable --now voice.service
```

`--now` starts a **stopped** unit. It does not restart a running one. So the script writes the new binary to `/usr/local/bin/voice`, leaves the old process running the old code, and prints:

```
Installed /usr/local/bin/voice with restricted agent account voice-agent.
```

The message is true and useless: the file was installed, and the program serving the household was not the one installed. Every reinstall since #15 has deployed a binary nobody then ran.

This is the same defect class as the four unfailable artifacts already found in this repo — a step that reports success for work that did not take effect. It is worse than those, because it silently invalidates any conclusion drawn from the deployed program's behaviour.

## The fix

1. Restart rather than merely enable, so the installed binary becomes the running one.
2. **Verify it**, rather than trusting the restart: compare the `sha256` of `/usr/local/bin/voice` against the resolved target of `/proc/<MainPID>/exe`, and fail the installer loudly if they differ.

Point 2 is the part that matters. A restart can be refused, a unit can fail and fall back, and a service can be masked; the installer must not report success unless the process actually running is the file it just installed.

## Acceptance criteria

- The installer fails, with a message naming both hashes, when the running executable is not the installed one.
- A test or check that distinguishes the two cases: installing while the service is **already running** must end with the new binary running. Asserting only that the file was written would pass the current, broken behaviour.
- The device confirms `stage=` lines appear after an install, which is the observable that was missing.

## Evidence

- `scripts/install-local.sh` final lines: `systemctl enable --now voice.service`
- Device: running PID continued emitting the pre-#15 format after a successful-looking install
- Device: `journalctl -u voice` shows no `stage=capture`, `stage=segment`, `stage=transcribe` or `stage=wake` lines in a window where the service transcribed several utterances

jerryfane:
Closes #24.

## Why the device served the household from a pre-#15 binary

`scripts/install-local.sh` ended with:

```sh
sudo systemctl enable --now voice.service
```

`--now` starts a **stopped** unit and leaves a running one alone. So the script wrote the new binary to `/usr/local/bin/voice`, the old process kept serving, and the script printed:

```
Installed /usr/local/bin/voice with restricted agent account voice-agent.
```

True and useless. The file was installed; the program answering the household was not the one installed. **Every install since the per-stage logging landed deployed a binary nobody then ran** — which is also why no `stage=` lines ever appeared in the journal, and why several of my own statements about the device's behaviour were about a program that is not in this repo.

## The fix

1. `systemctl restart` rather than `enable --now`.
2. **Verify**, rather than trust the restart: compare the `sha256` of the installed file against the resolved target of `/proc/<MainPID>/exe`, and fail — naming both hashes — when they differ.

Point 2 is the part that matters. A restart can be refused, a unit can fail and leave the old process up, a service can be masked. An installer that reports success without checking what is *running* is the same unfailable artifact as the `doctor` line that called a missing engine OK.

Success now names the evidence instead of implying it:

```
Verified pid 1121960 is running that exact binary (a4f45afd…).
```

## Verified three ways

Against a harness with a fake `systemctl` and a symlinked `/proc/<pid>/exe`:

| Case | Result |
|---|---|
| Running binary differs from installed | **fails**, printing both hashes |
| Running binary matches | passes |
| `MainPID=0` | **fails**: "voice.service is not running after restart" |

## The bug inside my own test

My first harness extracted the guard with `sed '/^installed_hash=/,/^fi$/p'`, which stops at the **first** `fi` — the `MainPID` check — so the hash comparison was never in the harness at all. It reported a pass for a check it had not run.

That is precisely the defect class this PR removes, and it nearly happened inside the test for it. Recorded because the pattern is now at nine instances in this repo, and the only reliable defence has been running the failing case deliberately rather than admiring a green result.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d654ec57df8166-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/voice PR #25 at exact head 184e30b44e5c62e047c123866b11cf67103a7ba0 (single commit, scripts/install-local.sh only, +32/-2). Build/vet/test all pass; sh -n clean. Independently built a fake-sudo/fake-systemctl/symlinked-/proc harness and ran the actual verification block (through its closing fi, not truncated) against 6 scenarios: matching hash passes, differing hash fails naming both hashes/paths, MainPID=0 fails, empty MainPID fails, simulated permission-denied sudo readlink fails closed, and a vanished pid directory fails closed via the 'cannot read' branch. Attacked the four angles requested: (1) wrapper ExecStart - resolved as a non-issue, packaging/voice.service:9 ExecStart is the bare binary (Type=simple), voice-agent-run is a separate script invoked by the running process for brain calls, not the unit's ExecStart, and grepping the Go source found no self-fork/re-exec, so /proc/MainPID/exe legitimately equals the installed file for every honest install; (2) ProtectSystem=strict/ProtectHome=read-only/NoNewPrivileges constrain the service's own mount namespace, not an external root's ability to read /proc/PID/exe (no PrivateUsers/PID-namespace isolation is set, and root retains CAP_SYS_PTRACE), and a forced permission-denied on sudo readlink was shown to fail closed with a clear message rather than silently pass - no legitimate-install path to that failure was found on this unit file; (3) Type=simple with no sd_notify anywhere in the Go source means MainPID is set synchronously at fork with no readiness-gated startup delay, so the requester's feared slow-start race does not apply here, and an immediate-crash case correctly reads MainPID=0 and fails; (4) the enable-plus-restart change to voice-auth-broker@ is consistent with the same staleness problem (voice-omp is reinstalled unconditionally every run too) but adds a real, non-blocking gap: unlike voice.service, the broker restart has no post-restart hash verification, and the forced bounce co
[truncated]

**Findings**
- **P3**
  - description: voice-auth-broker@ is restarted unconditionally on every install (like voice.service, since voice-omp is also reinstalled each run) but has no post-restart sha256/MainPID verification, so the exact staleness defect this PR fixes for voice.service could still recur silently for the broker if its restart is ever refused or masked.
  - evidence: scripts/install-local.sh:82-89 (enable/restart of voice-auth-broker@ has no verification block, contrast with lines 96-113 which only cover voice.service)
- **P3**
  - description: The 'cannot read /proc/$main_pid/exe' branch assumes readlink -f fails when the exe entry is missing, but GNU readlink -f tolerates a missing final path component (canonicalize-missing) and only fails on a missing intermediate directory; in the synthetic case of an existing pid dir with no exe entry, execution instead falls through to sudo sha256sum on a nonexistent path, leaking an unsuppressed stderr line before still failing closed via the empty-hash mismatch. Not reachable in practice for a real live process on Linux, and the install still correctly fails either way - noted as a diagnostics-clarity gap, not a correctness bug.
  - evidence: scripts/install-local.sh:102-113; reproduced live with a scratch harness (fake sudo/systemctl, /proc dir present without an exe symlink) which printed a stray 'sha256sum: ...: No such file or directory' to stderr but still exited 1 with the intended 'install failed' message

**Tests Run**
- CGO_ENABLED=0 go build ./... (clean)
- CGO_ENABLED=0 go vet ./... (clean)
- CGO_ENABLED=0 go test ./... (all packages pass: audio, codecheck, device, feedback, hid, session, speech, timer, vad, wake)
- sh -n scripts/install-local.sh (clean)
- Extracted scripts/install-local.sh lines 96-113 (the verification block, through its closing fi) into a scratch harness with fake sudo/systemctl and a symlinked fake /proc/<pid>/exe: matching hash -> exit 0; differing hash -> exit 1 printing both installed and running hash+path; systemctl reporting MainPID=0 -> exit 1 'not running after restart'; systemctl reporting empty MainPID -> exit 1 'not running after restart'; simulated permission-denied sudo readlink -> exit 1 'cannot read ... to confirm what is running'; nonexistent pid directory -> exit 1 via the same 'cannot read' branch

PR #23, opened 2026-09-18
fix(wake): match the wake phrase by sound, and ship a default that works

jerryfane:
Closes #22.

## What happened on the device

The owner said *"Hey Voice, what time is it"* and nothing happened. The journal:

```
heard="Hey boys." peak=0.2551
heard="What time is it?"
```

Voice heard him perfectly. The microphone worked, the transcription worked, and the wake phrase never matched — whisper renders *"Hey Voice"* as **"Hey boys"** for this speaker, and the matcher compared **spelling**.

`hey boys` vs `hey voice` is 4 edits over 9 characters: a normalised distance of **0.44**. No tolerance loose enough to accept that could still reject ordinary conversation, and the prefix requirement in `Match` exists precisely to prevent incidental activation.

## The fix

Reduce both transcript and phrase to a **sound skeleton**, then apply the configured tolerance to that:

- consonants folded into the classes recognisers confuse: `b/v/f/p`, `s/z/c/x`, `d/t`, `g/k/q`, `m/n`, `l/r`
- vowels kept but coarsened to front (`e/i/y`) and back (`a/o/u`)

## Two things I got wrong first, both caught before merge

**Dropping vowels entirely.** My first skeleton discarded them, which collapsed `hey buzz` onto `hey voice` — the matcher would have woken Voice during ordinary talk. Keeping two coarse vowel classes separates `boys` from `buzz` while keeping `boys` and `voice` neighbours. The mutant that drops vowels fails the rejection test on exactly that phrase:

```
--- FAIL: TestMatchRejectsOrdinarySpeech
    "hey buzz light year" matched "hey voice"; this would wake Voice during ordinary conversation
```

**Bypassing `fuzz`.** My first version accepted sound-alikes even at `fuzz = 0`, silently breaking the documented contract that `0` demands an exact transcript. The repo's own `TestExactModeRejectsApproximateWakePhrase` caught it. Sound matching is a tolerance, so the tolerance governs it.

I also asserted two skeleton pairs agreed (`voice`/`vois`, `sink`/`zinc`) that do not — my assumption about the folding rather than a fact about it. Corrected, with the reason recorded in the test.

## The shipped default was the other half of the defect

`wake.fuzz` defaulted to **0**, so out of the box a single wrong character defeated the wake phrase. It is `0.2` now.

The test asserts the **default wakes on the transcript recorded from the device**, not that the number is non-zero — any non-zero value would satisfy a number check, and this default's whole failure was that it looked reasonable while being unusable. With the default back at `0`:

[...]

PR #27, opened 2026-09-18
feat(session): time the planner and the speaker, the only unlogged stages

### LINKED ISSUE #26: The planner and speaker stages are unlogged, so the slowest part of an interaction is invisible
## Observed on the device

A real interaction took about a minute. The per-stage logging from PR #15 accounted for six seconds of it:

| Stage | Measured |
|---|---|
| transcribe (2.42 s of audio) | 2.6 s |
| transcribe (2.84 s of audio) | 2.8 s |
| wake match | instant |
| **everything after the command was understood** | **unaccounted** |

## Cause

`internal/session/session.go` logs `stage=capture`, `stage=segment`, `stage=transcribe` and `stage=wake`. The planner call (`HandleText`) and the speaker (`Speak`) log **nothing** on success — only `speak: %v` on error.

So the two stages that own almost the entire wait are the only ones nobody can see. The pipeline can be diagnosed everywhere except where it is slow, which is where diagnosis matters.

## The fix

Log `stage=plan` and `stage=speak` with durations, and route every `Speak` call through one timed helper so no callsite can be added later without timing.

- `stage=plan state=ok ms=… reply=…` — planner latency and reply length
- `stage=speak state=ok ms=… chars=…` — synthesis plus playback, which is what a person actually waits through, with the text length so a slow reply can be told from a long one
- failures carry the same durations rather than only an error

## Acceptance criteria

The test must require the **duration**, not merely the stage name: a stage line with no `ms=` cannot answer where an interaction went, and asserting only the stage name would pass a line that is useless for the purpose it exists for.

jerryfane:
Closes #26.

## Observed on the device

A real interaction took about a minute. The per-stage logging from PR #15 accounted for six seconds of it:

| Stage | Measured |
|---|---|
| transcribe, 2.42 s of audio | 2.6 s |
| transcribe, 2.84 s of audio | 2.8 s |
| wake match | instant |
| **after the command was understood** | **invisible** |

`HandleText` and `Speak` logged nothing on success — only `speak: %v` on error. So the pipeline could be diagnosed everywhere except where it was slow, which is where diagnosis matters.

## The change

- `stage=plan state=ok ms=… reply=…` — planner latency and reply length
- `stage=speak state=ok ms=… chars=…` — synthesis **plus** playback, which is what a person actually waits through, with text length so a slow reply can be told from a long one
- failures carry the same durations rather than only an error
- every `Speak` callsite goes through one timed helper, so a new one cannot be added without timing

## The test requires the duration

Not just the stage name. A stage line with no `ms=` cannot answer where an interaction went, so asserting the name alone would accept a line that is useless for its purpose.

Both mutants fail: removing the `stage=plan` line, and logging `stage=speak state=ok` without a duration.

## Note on provenance

This work was briefly swept into PR #23 because I switched branches mid-edit. It is split out here unchanged; #23 is the wake matcher alone.


jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d656b4ac7e6829-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d656b4ac7e6829-1`

**Summary:** Reviewed jerryfane/voice PR #27 at exact head 5e5c203e56edeeb19c589fc74bdb437a818d467d (base main 9d9b68a), which adds stage=plan and stage=speak timing logs. Build (CGO_ENABLED=0), vet, gofmt, full test suite, and errcheck (-excludeonly, matching CI) all pass; -race could not be run because this read-only review seat cannot read /usr/include for cgo compilation, confirmed as an environment-wide restriction unrelated to the code. Verified all five requested attack points: (1) Speak's timing is honest — CommandPlayer.PlayWAV blocks on cmd.Run(), so stage=speak covers synthesis+playback synchronously as claimed; (2) error behavior at all three converted Speak call sites is unchanged from main, only the log text differs; (3) sinceMS truncates toward zero and will realistically log ms=0 in the added test's instant fakes, which is cosmetically ambiguous but does not break the test or mislead about failure state (P3, non-blocking); (4) the test's substring 'stage=plan state=ok ms=' is placed adjacent to the actual %d in the log format, so a mutant dropping the duration token would fail it — confirmed by static reading since the worktree has no writable path anywhere to run the mutation live; (5) the two added log lines fire once per accepted command, negligible next to existing per-stage logging and to TTS/playback latency itself. One finding beyond the five: the PR's own claim that 'one timed helper for every Speak callsite' now exists is false — internal/session/timers.go's pre-existing say() helper calls a.Speak directly at 8 production sites (timer prompt/set/cancel/list, missed/fired timer announcements, and the post-planner-failure fallback) and still only logs bare 'speak: %v' on error, nothing on success, leaving that traffic exactly as unobservable as before this fix. No prior verdicts exist for PR #27 in GITMOOT_PRIOR_VERDICTS (as_of 2026-09-18T06:34:14Z), so there were no earlier findings to answer. Same-model-family caveat: this review was performed by Claude 
[truncated]
**Severity:** `P2`

**Findings**
- **say() bypasses the new timed Speak helper at 8 production call sites** (P2)
  - description: internal/session/timers.go:124-128 defines say(), a second Speak entry point used at timers.go:25,33,37,40,43,105,112 and session.go:217. It still only logs 'speak: %v' on failure and nothing on success, so timer prompts/confirmations/announcements and the planner-failure fallback message remain exactly as invisible to timing as before this PR — contradicting the PR's stated claim that every Speak callsite is now timed.
  - locator: internal/session/timers.go:124-128 (say definition); call sites timers.go:25,33,37,40,43,105,112; session.go:217
- **sinceMS truncates to 0ms for sub-millisecond stages** (P3)
  - description: time.Duration.Milliseconds() truncates toward zero, so a very fast plan or speak stage logs ms=0, which is indistinguishable at a glance from 'not measured.' Confirmed this occurs in the added unit test itself since its planner/TTS/player fakes are instant with no artificial delay; does not fail the test since it only checks for the literal 'ms=' substring, and is unlikely in production speak (which spawns a playback subprocess) but plausible for a very fast planner.
  - locator: internal/session/session.go: sinceMS function and its two call sites; internal/session/session_test.go:395-420 (countingPlanner/countingPlayer have no delay)

**Tests Run**
- CGO_ENABLED=0 go build ./... -> success
- CGO_ENABLED=0 go vet ./... -> clean, no output
- gofmt -l . -> clean, no output
- CGO_ENABLED=0 go test ./... -> ok for all packages including internal/session (2.802s), internal/speech (8.120s), internal/timer (3.650s)
- CGO_ENABLED=0 go test ./internal/session/... -run TestEachPipelineStageIsObservable -v -> PASS
- go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... -> exit 0, no output (zero discarded call results), matching CI's own errcheck step
- go test ./internal/session/... -race (with and without dangerouslyDisableSandbox) -> build failed: runtime/cgo fatal error reading /usr/include/stdc-predef.h: Permission denied; reproduced identically for plain `cat` on /usr/include/stdc-predef.h, /etc/hostname, /etc/bash.bashrc as root, confirming an environment-wide read restriction in this review seat rather than a code or repo issue

jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d657814b903cfe-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d657814b903cfe-1`

**Decision:** `approved`
**Summary:** At head a3dfb0e (one commit since the reviewer's prior head 5e5c203): f1 (P2, say() bypassing the timed Speak helper) is resolved — say() now calls speakLogged directly (timers.go:128-129), and a repo-wide grep confirms the only remaining a.Speak() call inside internal/session is inside speakLogged itself (session.go:230), so all 8 former call sites now inherit timing. f2 (P3, sinceMS truncating to ms=0) is resolved — sinceMS now computes fractional milliseconds from microseconds (session.go:241) and a new test proves this against instant fakes. Both are verified by a passing full test suite, clean vet/gofmt, and reasoning against the pre-PR code (the worktree is genuinely read-only, so mutants could not be physically applied, only analyzed). New P3: the %.3f format is unconditional, so the PR's own motivating case (a 7.102s planner call) now prints noisy 'ms=7102.000' instead of 'ms=7102', which reads worse for the common/important case that this instrumentation exists to serve. Also answering the requester's Q1 directly: the single-entry-point claim holds only within the interactive session/journal — cmd/voice/main.go:123,134 (the 'voice say' and 'voice ask' one-shot CLI subcommands) call a.Speak directly and remain untimed, but that code is unchanged by this PR and out of this round's scope, so it is reported as context, not a new defect. Same-model-family caveat: this review was run by Claude Sonnet 5, plausibly the same family that authored the PR's fixes.
**Severity:** `P3`

**Findings**
- **say() bypasses the new timed Speak helper at 8 production call sites** (P2)
  - continues_uid: jerryfane/voice#27-f1
  - detail: Resolved. timers.go:128-129 now has say() call a.speakLogged(ctx, text) instead of calling a.Speak directly and logging only 'speak: %v' on failure. Grep of internal/session for '.Speak(' shows the sole remaining call is inside speakLogged (session.go:230), so all 8 former call sites (timers.go:25,33,37,40,43,105,112; session.go:217) now emit 'stage=speak state=ok/failed ms=...'. Verified with `CGO_ENABLED=0 go test ./internal/session/... -run TestTimerSpeechIsTimedLikeEveryOtherReply -v` (PASS); by inspection this test would fail against the pre-PR say() implementation since that version never logged anything on success.
  - file: internal/session/timers.go:124-129
  - state: answered
- **sinceMS truncates to 0ms for sub-millisecond stages** (P3)
  - continues_uid: jerryfane/voice#27-f2
  - detail: Resolved. sinceMS (session.go:241) now returns time.Since(t).Microseconds() as float64 divided by 1000, and all four call sites use '%.3f' instead of '%d' (session.go:216,220,231,234). TestStageDurationsKeepSubMillisecondResolution asserts the logged line never renders a bare integer 'ms=0' when fakes are instant, and passes (`CGO_ENABLED=0 go test ./internal/session/... -run TestStageDurationsKeepSubMillisecondResolution -v`). By inspection this would have failed under the old int64 Milliseconds()+%d implementation.
  - file: internal/session/session.go:216-241
  - state: answered
- **ms=%.3f always renders three decimals, adding noise to the common multi-second case** (P3)
  - detail: The new fixed-precision format is unconditional: a 7.102s remote-planner call (the exact figure cited as this PR's motivating measurement) now logs 'ms=7102.000' instead of the prior 'ms=7102'. The fractional precision only matters for the rare sub-millisecond case that TestStageDurationsKeepSubMillisecondResolution targets; for the dominant case (multi-hundred-ms to multi-second plan/speak stages) it adds three redundant trailing zeros to every line, working against the journal-readability goal this instrumentation exists to serve. A format that shows decimals only when non-zero (e.g. %g) or trims trailing zeros would fix both the truncation-to-0 problem and this readability regression.
  - file: internal/session/session.go:216,220,231,234 (format strings) and :241 (sinceMS)

**Tests Run**
- CGO_ENABLED=0 go build ./... — success (plain `go build ./...` fails in this sandbox on a cgo header permission error unrelated to this PR's diff)
- gofmt -l . — clean, no output
- CGO_ENABLED=0 go test ./... — all packages ok, internal/session 2.929s
- CGO_ENABLED=0 go test ./internal/session/... -run 'TestTimerSpeechIsTimedLikeEveryOtherReply|TestStageDurationsKeepSubMillisecondResolution|TestEachPipelineStageIsObservable' -v — all PASS
- grep -rn '\.Speak(' --include='*.go' . — confirmed sole internal/session call at session.go:230 (inside speakLogged); two additional direct calls found at cmd/voice/main.go:123,134, unchanged by this PR
- git diff 5e5c203e56edeeb19c589fc74bdb437a818d467d..a3dfb0edf794ddd98a5336b53a73710e7cd448d2 -- internal/session/session.go internal/session/session_test.go internal/session/timers.go — reviewed in full
- git log --oneline 5e5c203..a3dfb0e — confirms exactly one commit (a3dfb0e) since prior head
- git log --oneline -- cmd/voice/main.go / git diff 9d9b68a..a3dfb0e -- cmd/voice/main.go — confirms main.go untouched by this PR

Issue #1, opened 2026-09-17
Roadmap: activation feedback, scheduling, Home Assistant, and memory controls


jerryfane:
## Goal

Make Voice feel responsive and useful for daily household routines while preserving its strict wake gate, local-first behavior, restricted agent boundary, and explicit control over stored memory.

## User experience

- Saying **“Hey Voice”** gives immediate physical feedback: the microphone light changes and a short, pleasant acknowledgement sound plays.
- Voice supports dependable timers, alarms, and reminders without requiring the LLM for execution or delivery.
- Voice integrates with Home Assistant and can discover compatible devices on the local network, including appliances such as an air fryer when they expose a supported local protocol or Home Assistant entity.
- Users can inspect, add, and remove durable memories.

## Sub-issues

- [ ] #2 — Activation feedback: microphone light and acknowledgement sound
- [ ] #3 — Local persistent timers
- [ ] #4 — Local alarms and recurrence
- [ ] #5 — Persistent spoken reminders
- [ ] #6 — Home Assistant integration and safe local-device discovery
- [ ] #7 — Memory inspection and deletion controls

These are also linked through GitHub’s parent/sub-issue relationship.

## Cross-cutting requirements

- Ambient noise and standalone activation feedback must never call the LLM.
- Scheduling must survive service restarts and work when the model or network is unavailable.
- Network discovery must be explicit, bounded, local-only, and non-destructive; do not perform unrestricted port scanning.
- Discovered devices must remain read-only until the user explicitly approves and names them.
- Home Assistant tokens and other secrets must not enter prompts, conversation history, logs, or `MEMORY.md`.
- Device actions must continue through typed, allowlisted capabilities rather than arbitrary model-generated network requests.
- All audible and visual feedback must respect Voice’s on/off state.
- New behavior needs observable tests and an installed-device smoke test.

## Suggested delivery order

1. Activation feedback
2. Timers
3. Alarms
4. Reminders
5. Home Assistant and discovery
6. Memory controls

Timers establish the scheduler used by alarms and reminders. Home Assistant discovery should reuse the existing validated device registry rather than exposing unrestricted APIs to the agent.

jerryfane:
## Roadmap status from the voice seat

| Workstream | State |
| --- | --- |
| Activation feedback (#2) | Implemented, PR #8, independently reviewed; on-device criterion **blocked on the owner**, not failed |
[...]

PR #28, opened 2026-09-18
fix(hid): log the exact report bytes, because a dark light proves nothing without them

jerryfane:
Refs #2 — this is instrumentation for that issue's unsolved on-device criterion, not a fix for it.

Note on naming: the branch is `issue-28-log-report-bytes`, which implies an issue #28 that does not exist — GitHub numbered this PR 28 and my branch name was a guess made before the PR existed. The work belongs to #2. Recorded rather than silently renamed, since a branch name implying a missing issue is how provenance gets lost.

## Observed on the device

| Path | Result |
|---|---|
| Raw report `02 09 00` written directly | **solid red** |
| Voice's own path with `feedback.light=hold` | **no visible change, no error logged** |

The LED works. The code reports success. Nothing lights up.

## Why it could not be diagnosed

The device seat named it precisely: a successful report write logged **nothing**, so a light that did not come on was indistinguishable from

1. a write that never happened,
2. a write of the wrong bytes, or
3. firmware ignoring a correct report.

Three different causes, one observable. That is the same defect class as every other one found in this repo today, applied to the one path that had no instrumentation at all.

## The change

Every report write now logs its exact bytes, its **length**, the report ID and the bit it set:

```
stage=hid state=written path=/dev/hidraw5 report=2 bytes=02 09 00 len=3 bit=byte0/mask0x08 on=true
```

The length is evidence, not a detail: `internal/hid/telephony.go` already documents that firmware treats a report of the wrong length as a different report, which makes a short report a prime suspect for a silent failure of exactly this shape.

No behaviour change. The logger is optional and nil-safe, and the write path is untouched apart from one line after a successful close.

## Tests

- the bytes and the length must both appear
- a second bit written into the same report must carry the **retained** off-hook bit, so a report sending `08` where `09` is required would be caught — that is the leading hypothesis for the missing light
- the mutant that logs without bytes or length fails all three assertions

## What this does not do

It does not fix the light. It makes the next run say which of the three causes it is. Issue #2 stays open; fixing before measuring is how several wrong turns happened today.


jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d657def9892350-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d657def9892350-1`

**Summary:** Reviewed jerryfane/voice PR #28 at exact head b36806616e5a207fc762991c01ba4c61b31b8992 (adds per-write hid.Telephony logging of exact report bytes/length/bit, wired via SetLogger in session/build.go). Build (CGO_ENABLED=0, matching CI), vet, gofmt, full test suite, and errcheck all reproduced green. Independently traced the diagnostic claim end to end: the new log line, combined with two pre-existing untouched log sites (build.go's Nop-reason 'wake light' log and feedback.go's write-error log), does separate the three hypotheses the operator named (write never happened / wrong bytes / firmware ignored a correct report) by elimination — but only as a system, not from this one line alone. Confirmed by static trace that logf() never locks t.mu, so no self-deadlock exists even though write() calls it while holding the mutex; confirmed by code trace that off-hook writes fire once per process lifetime (not per capture cycle) so volume is not a concern; confirmed bytes=% x renders as expected (02 09 00) with no grep-separator ambiguity. However, mutation-testing the new test empirically found that its retained-bit assertion (strings.Contains(wire[1], "09")) is not scoped to the payload bytes: because the log line embeds path=t.TempDir(), whose random suffix frequently contains the substring '09' by chance, a deliberately corrupted payload byte (0x01 instead of 0x09, i.e. the real regression this test exists to catch) still passed in 136/2000 (6.8%) runs. This does not affect the production log line itself (real device paths are static, not random tempdirs), so the shipped diagnostic mechanism is sound and does answer the operator's question, but the test that is supposed to guard it is unreliable for its stated purpose and should be fixed before being trusted as a regression gate. No prior verdicts exist for PR #28 in GITMOOT_PRIOR_VERDICTS (as_of 2026-09-18T06:55:26Z), so no prior findings to answer. Same-model-family caveat: this review was performed by a Claude model (S
[truncated]
**Severity:** `P2`

**Findings**
- **Test assertion can pass for the wrong reason: matches t.TempDir()'s random path suffix, not the payload byte** (P2)
  - detail: internal/hid/telephony_test.go:275 checks strings.Contains(wire[1], "09") against the full log line, which includes path=<t.TempDir()>. The OS-assigned random tempdir suffix frequently contains the literal substring '09' by chance. Reproduced empirically: mutated a disposable copy so the actual payload byte written was 0x01 instead of 0x09 (simulating the hold bit never reaching the wire — the real regression this test exists to catch), then ran `go test -run TestWriteLogsTheExactReportBytesAndLength -count=2000 -v`: 136/2000 runs (6.8%) passed anyway, matching '09' inside the random path rather than the corrupted bytes field. Does not affect production behavior (real device paths are static config strings, not random tempdirs), but undermines confidence that this test reliably guards the property it claims to prove. Recommend anchoring on the bytes= field specifically, e.g. strings.Contains(wire[1], "bytes=02 09 00") or extracting the substring after "bytes=" before matching.
  - evidence_locator: internal/hid/telephony_test.go:275-277; reproduced via disposable copy + `go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=2000 -v`
  - id: retained-bit-assertion-flaky
- **The new log line alone does not distinguish 'write never happened' — it relies on two pre-existing, untouched log sites** (P3)
  - detail: logf() in write() only fires on success. 'Write never happened' is instead caught by session/build.go:84-87 ('wake light: %s', fired when NewLight falls back to Nop, e.g. from Open's descriptor-parse-failure/StandardOffHook fallback) and feedback/feedback.go:94 ('feedback indicator %s: %v', fired on a syscall error from write()). This is not a defect — the combination does work — but it's worth the operator knowing the new line is not self-sufficient in isolation; the engineer must correlate it with those two other log lines to fully separate all three hypotheses.
  - evidence_locator: internal/session/build.go:84-87; internal/feedback/feedback.go:90-96; internal/hid/telephony.go:194-227
  - id: diagnostic-line-needs-existing-logs

**Tests Run**
- CGO_ENABLED=0 go build ./... — success
- CGO_ENABLED=0 go vet ./... — clean
- gofmt -l . — clean (no files listed)
- CGO_ENABLED=0 go test ./internal/hid/... -v — all 12 tests + fuzz seeds PASS, including new TestWriteLogsTheExactReportBytesAndLength
- CGO_ENABLED=0 go test ./... — full suite green (11 packages ok)
- CGO_ENABLED=0 go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... — exit 0, no findings (matches CI step exactly)
- CGO_ENABLED=0 go test ./internal/codecheck/... -v — all discard-checker tests pass, 43 files scanned, none flagged
- Mutant #1 (disposable copy): removed bytes=/len=/on= from the logf call in write() — TestWriteLogsTheExactReportBytesAndLength fails all 3 format assertions plus the retained-bit check, matching the PR's claimed mutant-kill evidence
- Mutant #2 (disposable copy): corrupted the actual payload byte (index 1) to 0x01 while keeping bytes=/len=/on= present, then ran `go test -run TestWriteLogsTheExactReportBytesAndLength -count=2000 -v` — 1864/2000 correctly failed, but 136/2000 (6.8%) spuriously passed via the '09' substring appearing in t.TempDir()'s random path segment
- go run scratch harness verifying bytes=% x formatting for []byte{2,9,0}, {2,0,0}, {2,255,16} renders as '02 09 00', '02 00 00', '02 ff 10'
- Static trace of telephony.go SetLogger/logf/write mutex usage — confirmed logf() never locks t.mu, ruling out same-goroutine self-deadlock; corroborated by hid test suite completing without hanging while exercising this exact code path
- Static trace of write() call frequency via internal/audio/command.go (setOffHook) and internal/session/session.go (Run calling Stream once) — confirmed off-hook writes fire once per process lifetime, not per capture cycle
- Checked GITMOOT_PRIOR_VERDICTS (as_of 2026-09-18T06:55:26Z) — no entries for pull_request 28

jerryfane:
## Both findings answered at head `f60728b9784805444f0813371675fc12a8a544f8`

### `#28-f1` (P2) — my test could pass for the wrong reason

This is the finding of the day, because it is the pattern *inside the test written to prevent it*. The assertion matched `"09"` anywhere in the log line, and `t.TempDir()` puts a random number in `path=`. You mutated the payload so the hold bit never reached the wire — the exact regression the test exists to catch — and **136 of 2000 runs passed anyway**, matching the path instead of the payload.

Fixed: the assertion now reads the `bytes=` field through a helper and compares it whole. Re-running your mutant (`payload[bit.Byte] = 0x01`) with the corrected test:

```
50 runs, 50 failures
second report carried "02 01 00", want "02 09 00"
```

Every run fails, and the failure names what actually went on the wire.

### `#28-f2` (P3) — the line is not self-sufficient

You were right that this is not a defect but matters to whoever reads it: the write log fires only on success, so it cannot separate *wrote the wrong bytes* from *never wrote at all* — the second case leaves no line, and catching it required knowing to correlate `wake light: %s` in `build.go` and `feedback indicator %s: %v` in `feedback.go`.

`Telephony.LogReady` now records the path and capabilities when the device opens, so **a ready line with no write lines after it is a diagnosis rather than a gap**.

It lives in `hid` rather than in `build` specifically so it can be tested: the assertion requires the path and a capability string, and requires that a nil device neither panics nor claims to be ready. The mutant that logs `state=ready` with neither fails it.

## One thing I removed

I had written a session-level test that asserted only that `Build` returned a non-nil feedback. It proved nothing about the ready line and would have padded the suite, so I deleted it rather than leaving it to look like coverage.

## Verification at this head

`go test ./...` green; `CGO_ENABLED=1 go test -race ./internal/hid/ ./internal/session/` green; `go vet`, `gofmt -l` clean; exclusion-free `errcheck` exits 0.

Still not a fix for the dark light — it is the measurement that will say which of the three causes it is.

jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d658af084cce3d-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d658af084cce3d-1`

**Decision:** `approved`
**Summary:** F1 is fixed and verified by re-running the exact mutation that previously passed 6.8% of the time: cloned the repo at f60728b, mutated write() so the hold bit (mask 0x08) never reaches the payload, and ran the new bytesField-based assertion 200/200 times — 200/200 FAIL, reproducing the operator's cited wire value '02 01 00' exactly. F2 (LogReady) is structurally correct: it's called on every path where hid.Open returns a non-nil handle (including the descriptor-parse-failure fallback, where Describe()'s nil-safe Outputs.Bit correctly renders 'no known output controls' rather than a blank field), and skipped only when Open returns nil (device path doesn't exist), where there's nothing to report readiness for and the preceding error log already explains why. A third pre-existing log site (session.go:126, 'stage=capture state=failed') further helps separate write failures from silence, beyond the two sites F2 originally named. Two new, non-blocking P3s surfaced this round: the nil-device sub-assertion in TestLogReadyRecordsThePathAndCapabilities is decorative (confirmed by removing SetLogger's nil-guard: the test then fails via a nil-pointer panic on mu.Lock(), never via the 'nil device logged a ready line' callback, because that callback can never be stored on a genuinely nil receiver) — matches the operator's own instinct that it should probably be deleted. And the failing assertion's diagnostic message at telephony_test.go:283 prints the actual full log line as the 'want' value instead of the literal '02 09 00', which would confuse a future reader of a failure. Build, vet, gofmt, and the full non-race test suite are all clean at this head. No P1/P2 defects found in the changed code; approving.

**Findings**
- **Bytes-field assertion fix verified by mutation testing: flakiness is closed** (P2)
  - continues_uid: jerryfane/voice#28-f1
  - detail: Cloned the repo at exact head f60728b into a disposable scratch directory (outside the reviewed worktree) and patched internal/hid/telephony.go's write() so that when on=true and bit.Mask==0x08 (the hold bit), the mask is never OR'd into the payload — simulating the exact regression the test exists to catch. Ran `go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=200`: 200/200 runs FAIL, and the failure message correctly names the corrupted payload '02 01 00' (matching the operator's own reported mutant output exactly). This is a qualitative change from the prior head, where the same class of mutation passed 136/2000 (6.8%) by accident via the tempdir path. The fix (extracting the bytes= field via a dedicated helper and comparing the whole string) closes the flakiness. Reverted the mutation and confirmed the suite is green again before proceeding.
  - evidence_locator: internal/hid/telephony_test.go:275-283 (bytesField helper and equality check); mutation + `go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=200` in a disposable clone at f60728b, 200/200 FAIL
  - state: answered
- **LogReady closes most of the ambiguity, but 'ready + silence' still has two readings the journal can't tell apart** (P3)
  - continues_uid: jerryfane/voice#28-f2
  - detail: Traced every call site: LogReady() fires whenever c.Input.TelephonyHID is set and hid.Open returns a non-nil *Telephony (session/build.go:29-48), including the descriptor-parse-failure fallback where t.outs is nil — verified directly that Describe() on such a handle returns 'path (no known output controls)' rather than an empty caps field, because Outputs.Bit has a nil-safe receiver (internal/hid/descriptor.go:62-29). LogReady is a no-op only when Open itself returned nil (os.Stat failure on the configured path), and in that case the preceding `logger.Printf("telephony HID: %v", err)` already explains there is no device — so no false 'ready' claim is emitted. Also found a THIRD pre-existing log site beyond the two named in the prior finding: internal/session/session.go:126 emits 'stage=capture state=failed err=%v' whenever setOffHook's write fails and propagates through Stream's error channel, which helps separate an active write failure from a still-missing feature call. However, one ambiguous case remains uncovered by any of the four sites (ready-line, write-line, build.go:84 wake-light, feedback.go:95, session.go:126): 'ready line present, followed by no write lines and no error lines at all' is consistent with two different states — 'the feature hasn't had a reason to write yet' (benign) and 'a code path silently skips calling Set/SetReportBit without erroring' (a real bug) — and the journal alone cannot distinguish them. This isn't introduced by this PR; it's a residual property of 'absence of a log line proves only its own absence,' but it means the three-hypothesis separability claim doesn't fully cover a fourth practical case.
  - evidence_locator: internal/session/build.go:27-49; internal/hid/telephony.go:110-136,192-197; internal/hid/descriptor.go:23-29; internal/session/session.go:118-126; internal/feedback/feedback.go:90-96
  - state: answered
- **Nil-device sub-assertion in TestLogReadyRecordsThePathAndCapabilities is decorative: it cannot fail via the path it claims to guard** (P3)
  - detail: The test does `none.SetLogger(func(...){ t.Error("nil device logged a ready line") }); none.LogReady()` on a `var none *Telephony`. Verified by removing the `if t == nil { return }` guard from SetLogger in a disposable clone and re-running: the test then fails, but via a nil-pointer-dereference panic on `t.mu.Lock()` at telephony.go:179 — never via the injected t.Error callback. This is structural: SetLogger's nil-guard returns before `t.log = f` ever executes, so the closure is never stored anywhere a genuinely nil *Telephony could later read it from; LogReady's own nil-guard (or logf's) makes the callback unreachable regardless of which of the three nil-checks in the chain is present. The test does still catch a real regression (removing all nil-guards causes a panic instead of a clean no-op), but the specific 'must not log a ready line' assertion inside the callback can never be exercised by any code change short of restructuring Telephony's field storage. Matches the operator's own suspicion; recommend replacing it with a bare no-panic check (or deleting it) rather than keeping code that implies it verifies something it structurally cannot.
  - evidence_locator: internal/hid/telephony_test.go:326-331; reproduced by removing SetLogger's nil-guard in a disposable clone of f60728b and re-running `go test ./internal/hid/... -run TestLogReadyRecordsThePathAndCapabilities -v -count=1` — failed via panic at telephony.go:179, not via the injected t.Error
  - file: internal/hid/telephony_test.go
- **Failure message on the fixed bytes-field assertion prints the actual line as 'want', not the expected literal** (P3)
  - detail: `t.Errorf("second report carried %q, want %q (...)", got, wire[1])` passes `wire[1]` (the full actual log line) as the 'want' argument instead of the literal "02 09 00" that's actually being compared against two lines above. Confirmed live: forcing the mutation from the F1 check produces the message '... carried "02 01 00", want "stage=hid state=written path=... bytes=02 01 00 len=3 ..." (the retained off-hook bit plus hold)' — the printed 'want' is just a restatement of what was carried, not what was wanted. Doesn't affect the correctness of the comparison itself (which is fine), only the diagnostic a future engineer reads on a real failure.
  - evidence_locator: internal/hid/telephony_test.go:283; reproduced via the same mutation clone used for the F1 check, `go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -v -count=1`
  - file: internal/hid/telephony_test.go
- **bytesField's parse is safe today only because hex-pair bytes can't contain the anchor substring — and breaks loud, not silent, if that stops holding** (P3)
  - detail: bytesField locates "bytes=" then takes everything up to the first " len=". Today this is safe because the bytes field is strictly hex-pair-and-space output from "% x" formatting, which cannot contain the letters needed to spell " len=" (no 'l' in hex). Traced what happens if a future format change broke that assumption (e.g. inserting a new field between bytes= and len= whose own text happened to contain " len=", or reordering fields): bytesField would return a truncated/wrong substring, but since the test does a strict `!=` comparison against a hardcoded literal, the result is a spurious FAIL (loud), not a spurious PASS (silent) — the opposite failure direction from the original F1 bug, which was dangerous specifically because it passed for the wrong reason. So this is a real but low-severity fragility (anchored on field order/format, not a named delimiter), and any breakage would surface immediately as a failing test rather than reintroducing the silent-pass trap.
  - evidence_locator: internal/hid/telephony_test.go:288-301 (bytesField); internal/hid/telephony.go:236-237 (log format string it parses)
  - file: internal/hid/telephony_test.go

**Tests Run**
- git diff b368066 f60728b --stat / full diff — scoped review to the 3 declared changed files
- CGO_ENABLED=0 go build ./... — success
- gofmt -l . — clean, no files listed
- CGO_ENABLED=0 go test ./... — full suite green (all 12 packages ok, including internal/hid and internal/session)
- CGO_ENABLED=0 go test ./internal/hid/... -run 'TestWriteLogsTheExactReportBytesAndLength|TestLogReadyRecordsThePathAndCapabilities' -v -count=1 — both PASS
- go test ./internal/hid/... -race (also retried with dangerouslyDisableSandbox) — FAILED TO BUILD: cgo unavailable in this sandbox ('/usr/include/stdc-predef.h: Permission denied'); race detector could not be exercised here, so the PR's claimed 'race green' result was not independently reproduced by me
- Mutation test (F1): disposable clone of repo at f60728b outside the review worktree; patched write() so mask 0x08 (hold bit) is never applied when on=true; go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=200 — 200/200 FAIL, reproducing wire bytes '02 01 00'; reverted, suite green again
- Mutation test (nil-assertion): same clone; removed SetLogger's `if t == nil` guard; go test ./internal/hid/... -run TestLogReadyRecordsThePathAndCapabilities -v -count=1 — FAIL via nil-pointer panic on t.mu.Lock() at telephony.go:179, not via the injected t.Error callback; reverted
- Ad-hoc scratch test: constructed Telephony{outs: nil} and called Describe()/LogReady() in a disposable clone — Describe() returned 'path (no known output controls)', confirming the descriptor-parse-failure fallback still logs a meaningful capabilities field
**Diagnostics:** workflow advance failed: review approval held: no observed pull request row records a current head for jerryfane/voice#28, so this approval cannot be shown to be bound to it

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d658df7d3a0092-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/voice#28 at exact head df30df24761d16ebb367a82111b9e70fdb90c0e7, diffed against the prior-seen head b368066 (2 commits, 3 files: internal/hid/telephony.go, internal/hid/telephony_test.go, internal/session/build.go). Independently built (CGO_ENABLED=0 go build), vetted, gofmt-checked, ran the full suite, and reproduced both named mutants myself in isolated copies (not the reviewed tree). F1 (flaky retained-bit assertion) is fixed: bytesField() now extracts the bytes= field by exact boundary (Index 'bytes=' ... ' len=') and compares it with exact string equality, not Contains over the whole line. I independently authored the same class of mutant described in the original finding (drop the hold bit so the wire carries only the off-hook byte) and got 50/50 failures naming '02 01 00' vs '02 09 00' — matching the operator's reported numbers exactly. F2 (diagnostic line needs pre-existing logs) is narrowed but not closed: the new TestLogReadyRecordsThePathAndCapabilities now proves, in one stream, that 'ready, no writes' and 'ready, then a write' are distinguishable purely from LogReady+write()'s own log lines — that part of the ambiguity is resolved. What remains dependent on correlation is only the case of an attempted write that fails at the syscall level (open/write/close error), which leaves no state=written line and reads identically to 'never called Set' in the new fields alone. I found this is actually covered by a THIRD pre-existing log site the original finding didn't name — internal/session/session.go:124-125 ('stage=capture state=failed err=%v', fired when Recorder.Stream's setOffHook fails at capture start) — alongside the two already cited (build.go:84-87 wake-light fallback, feedback.go:94 indicator error). So the correlation surface is three sites, not two, but the underlying shape of F2 (the new line is not self-sufficient in isolation for that one failure mode) still holds. Answering the five attack points: (1) the three named hypothese
[truncated]
**Severity:** `P3`

**Findings**
- **Retained-bit assertion flakiness — fixed and independently verified** (P2)
  - continues_uid: jerryfane/voice#28-f1
  - detail: internal/hid/telephony_test.go now extracts the bytes= field via a bounded helper (Index 'bytes=' then ' len=') and asserts exact string equality against '02 09 00', replacing the old strings.Contains(wire[1], "09") over the whole line (which matched t.TempDir()'s random path suffix 6.8% of the time on a broken payload). I independently reproduced the same class of mutant in an isolated copy (dropped the hold bit in write() so retained state only carries the off-hook byte) and ran `go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=50 -v`: 50/50 failed, naming the exact corrupted wire value '02 01 00' vs expected '02 09 00' — matching the operator's reported reproduction precisely. Also ran the fixed test unmutated at -count=500: 500/500 pass, no residual flakiness observed.
  - evidence_locator: internal/hid/telephony_test.go:249-283 (bytesField, assertion); independently reproduced via isolated copy with write()'s bit-application mutated, `go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=50 -v` -> 50/50 FAIL with '02 01 00'
  - file: internal/hid/telephony_test.go
  - state: answered
  - uid: jerryfane/voice#28-f1
- **Diagnostic line still depends on correlation with pre-existing logs for one of three hypotheses, but the surface is now three sites, not two, and the ready/write pairing itself is directly tested** (P3)
  - continues_uid: jerryfane/voice#28-f2
  - detail: TestLogReadyRecordsThePathAndCapabilities now drives one logger through LogReady() then a successful Set(), proving 'ready with no write lines' and 'ready followed by a written line' are distinguishable purely from these two new log call sites (independently reproduced: mutating LogReady() to a no-op fails this test 20/20 in an isolated copy). What still requires correlation with other log sites is only the third hypothesis: an attempted write that fails at the syscall level (open/write/close error in write()) leaves no state=written line, so it reads identically to 'Set was never called' using only the new fields. I found this is resolvable via not two but three existing untouched log sites: internal/session/build.go:84-87 (wake-light fallback, fires at Open time only), internal/feedback/feedback.go:94 (indicator Set() error), and internal/session/session.go:124-125 ('stage=capture state=failed err=%v', fires when the capture path's setOffHook fails at stream start) — the last of these was not named in the original finding. None of this is a defect; it's a documentation/operator-knowledge gap that a reader must know to correlate three sites rather than read the new line in isolation.
  - evidence_locator: internal/hid/telephony.go:187-197 (LogReady), 216-227 (write's logf on success only); internal/session/build.go:84-87; internal/feedback/feedback.go:90-96; internal/session/session.go:103-129 (recErr handling, stage=capture state=failed at line 124-125); internal/hid/telephony_test.go:305-347 (TestLogReadyRecordsThePathAndCapabilities); independently reproduced LogReady-emits-nothing mutant via isolated copy, `go test ./internal/hid/... -run TestLogReadyRecordsThePathAndCapabilities -count=20 -v` -> 20/20 FAIL
  - file: internal/hid/telephony.go
  - state: answered
  - uid: jerryfane/voice#28-f2
- **New nil-safety sub-test only verifies 'does not panic', not its stated 'must not claim to be ready' half** (P3)
  - detail: TestLogReadyRecordsThePathAndCapabilities's nil-device block (internal/hid/telephony_test.go:330-332) registers a logger callback via none.SetLogger(...) that calls t.Error if invoked, then calls none.LogReady(), with a comment claiming this proves a nil *Telephony 'must not panic, and must not claim to be ready either.' But SetLogger itself returns immediately on a nil receiver before storing the callback (internal/hid/telephony.go:184-190), so the callback can never be invoked regardless of what LogReady's body does — the only property this sub-block actually exercises is the absence of a panic (which IS real coverage, since any field access on a nil *Telephony would panic and fail the test). I built a mutant where LogReady on a nil receiver routes through a package-level hook instead of t.log (bypassing t entirely, so no field is dereferenced and no panic occurs) and it passed 20/20 undetected in an isolated copy, confirming the 'must not claim to be ready' half of the stated intent is not actually checked. This is narrow and does not affect production correctness — no real construction path in this codebase produces a nil *Telephony that then has LogReady called on it outside this defensive guard — but it is exactly the shape of 'assertion that cannot fail for the reason it claims to guard' the operator asked me to hunt for.
  - evidence_locator: internal/hid/telephony_test.go:328-332; internal/hid/telephony.go:184-197 (SetLogger, LogReady nil guards); independently reproduced via isolated copy routing nil-receiver LogReady through a package-level hook instead of t.log, `go test ./internal/hid/... -run TestLogReadyRecordsThePathAndCapabilities -count=20 -v` -> 20/20 PASS despite the bypass
  - file: internal/hid/telephony_test.go
  - state: answered

**Tests Run**
- CGO_ENABLED=0 go build ./... (workspace at df30df2) -> success; plain `go build ./...` fails only on a sandbox permission error reading /usr/include/stdc-predef.h for cgo, unrelated to the PR
- CGO_ENABLED=0 go vet ./... -> clean, no output
- gofmt -l . -> clean, no output (no files listed)
- CGO_ENABLED=0 go test ./... -> all packages ok (audio, codecheck, device, feedback, hid, session, speech, timer, vad, wake)
- CGO_ENABLED=0 go test ./internal/hid/... -run 'TestWriteLogsTheExactReportBytesAndLength|TestLogReadyRecordsThePathAndCapabilities' -count=500 -v -> 500/500 pass, no flakiness
- Isolated copy (dropped hold-bit retention in write() to reproduce the exact prior mutant): go test ./internal/hid/... -run TestWriteLogsTheExactReportBytesAndLength -count=50 -v -> 50/50 FAIL, wire bytes '02 01 00' vs expected '02 09 00'
- Isolated copy (LogReady() short-circuited to a no-op): go test ./internal/hid/... -run TestLogReadyRecordsThePathAndCapabilities -count=20 -v -> 20/20 FAIL ('want one ready line, got 0')
- Isolated copy standalone check: Telephony{outs: nil}.LogReady() -> logged 'stage=hid state=ready path=/dev/hidraw-broken-descriptor caps=/dev/hidraw-broken-descriptor (no known output controls)', confirming graceful behavior on the descriptor-parse-failure path
- Isolated copy (nil-receiver LogReady routed through a package-level hook instead of t.log, bypassing the SetLogger no-op): go test ./internal/hid/... -run TestLogReadyRecordsThePathAndCapabilities -count=20 -v -> 20/20 PASS undetected, confirming the nil-safety sub-block only guards against a panic
**Diagnostics:** workflow advance failed: review approval held: no observed pull request row records a current head for jerryfane/voice#28, so this approval cannot be shown to be bound to it

jerryfane:
## `#28-f6` deferred to issue #29, with the reason

You marked it answered, and it is — but it is also a real defect of the family this PR is about, so I am not letting it pass as a comment quibble.

`SetLogger` returns early when the receiver is nil, so the callback is never installed and the `t.Error` inside it is **unreachable**. The comment claims the test proves a nil device does not claim to be ready; the code proves only that it does not panic. A can't-fail assertion, written during a round about can't-fail assertions.

Deferred rather than fixed here for the reason `#15-f4` and `#19-f10` were deferred: this PR is approved at `df30df24761d16ebb367a82111b9e70fdb90c0e7`, the merged head must equal the reviewed head, and #25 and #23 are waiting to reach a device that currently depends on a temporary `hey boys` alias. Trading that for a dead line in a test is the wrong order.

Issue #29 carries the fix and an acceptance criterion that a mutant removing the `t == nil` guard must break the test — or, failing that, the unprovable half and its comment get deleted, because an assertion that cannot fail reads as coverage while providing none.

Issue #4, opened 2026-09-17
Scheduling: local alarms and recurrence


jerryfane:
Parent: #1
Depends on: #3

## Goal

Add clock-based alarms on top of the persistent local scheduler established for timers.

## Supported requests

- “Hey Voice, wake me at seven tomorrow morning.”
- “Hey Voice, set an alarm for 6:30 every weekday.”
- “Hey Voice, list my alarms.”
- “Hey Voice, cancel tomorrow’s seven o’clock alarm.”
- “Hey Voice, snooze for ten minutes.”
- “Hey Voice, stop the alarm.”

## Design

- Reuse the timer scheduler and SQLite store rather than introducing a second clock loop.
- Store timezone-aware alarm definitions and compute the next UTC occurrence explicitly.
- Support one-shot alarms and a small, well-defined recurrence model: daily, weekdays, weekends, and selected weekdays.
- Execute and ring locally without requiring an LLM or network.
- Expose CLI operations for environments where voice input is off.

## Edge cases

- Daylight-saving gaps and repeated local times.
- Timezone changes after an alarm is created.
- Reboot, suspend, or service downtime spanning an occurrence.
- Snooze persistence and duplicate prevention.
- Multiple alarms ringing together.

## Acceptance criteria

- [ ] Create, list, enable, disable, cancel, stop, and snooze alarms.
- [ ] One-shot and recurring alarms survive restart/reboot.
- [ ] Common alarm operations use the local intent path.
- [ ] Timezone and DST behavior is documented and deterministically tested.
- [ ] Missed-alarm policy is configurable and never causes duplicate ringing.
- [ ] Alarm sound, volume/ramp, snooze duration, and maximum ring duration are configurable.
- [ ] Installed-service smoke test covers one-shot fire, snooze, and cancellation.

jerryfane:
## Semantics worked out ahead of implementation, and what the timer scheduler already gives you

Written under a coordinator hold: no code, no PR. This is the design record for when #4 starts, based on the scheduler PR #11 introduces for #3.

### What the existing scheduler satisfies unchanged

The scheduler is deliberately job-source-agnostic: it holds a set of items with absolute deadlines, sleeps until the earliest, persists after every change, and reports items that came due while the process was down. An alarm is a job with a deadline, so these need no change at all:

- one-shot alarms, including surviving restart and reboot;
- concurrent alarms ringing together (the fired channel already delivers several, and the session loop serialises the announcements);
[...]

Issue #3, opened 2026-09-17
Scheduling: local persistent timers


jerryfane:
Parent: #1

## Goal

Provide low-latency timers that execute locally and remain dependable without the LLM or network.

## Supported requests

- “Hey Voice, set a timer for twelve minutes.”
- “Hey Voice, set a pasta timer for eight minutes.”
- “Hey Voice, how long is left on the pasta timer?”
- “Hey Voice, list my timers.”
- “Hey Voice, cancel the pasta timer.”
- “Hey Voice, stop the timer.”

## Design

- Add a shared SQLite-backed scheduler owned by `voice-agent`.
- Parse common timer intents locally before the planner; ambiguous language may use the agent to produce a typed timer request, but scheduling and execution remain local.
- Use wall-clock expiry persisted in UTC so active timers recover correctly after service restart or reboot.
- Support multiple concurrent timers with stable IDs and optional user names.
- Ring through the configured speaker until acknowledged or until a configurable maximum duration.
- Keep timer state separate from conversational memory.

## Edge cases

- Reject zero, negative, and unreasonably long durations with a clear response.
- Define behavior for timers that expire while Voice is off or the machine is suspended.
- Avoid duplicate delivery after restart.
- Resolve ambiguous cancellation when multiple timers exist.

## Acceptance criteria

- [ ] Create, list, query, cancel, and stop timers by voice and CLI.
- [ ] Common timer operations do not invoke the LLM.
- [ ] Multiple named and unnamed timers work concurrently.
- [ ] Timers survive service restart and system reboot.
- [ ] Expiry plays a configurable local sound and identifies named timers.
- [ ] Expired/cancelled timers cannot fire twice.
- [ ] SQLite transitions and restart recovery have deterministic tests.
- [ ] Installed-service smoke test covers create, restart, fire, and acknowledge.

jerryfane:
## PR #11 covers part of this, and I owe a correction plus one decision

**Correction:** I filed #10 as a duplicate of this issue without checking for an existing sub-issue first. #10 is closed as a duplicate; this issue is the one of record, and PR #11 has been retargeted here.

### What PR #11 already satisfies

| Criterion | State |
| --- | --- |
| Common timer operations do not invoke the LLM | Done — verified with a brain command that logs every invocation: zero calls for set, query and cancel |
| Multiple concurrent timers | Done — fire in deadline order, individually cancellable |
[...]

PR #31, opened 2026-09-18
fix(brain): answer the clock locally, and play a thinking sound while the model works

### LINKED ISSUE #30: The local no-model fast path ships disabled, so "what time is it" costs a 7-second model call
## Observed on the device

The owner asked *"what time is it"* and waited about 14 seconds, of which **7.102 s** was a remote model call. The Pi asked a language model across the internet what time it was, while holding the answer in its own clock.

## Cause

`internal/brain/rules.go` already answers time, date, stop/cancel and device commands locally, with no model round-trip. It is gated on `brain.rules`, and `config.Default()` ships:

```go
Rules:   false,
```

The comment above the field reads *"Rules enables the local no-model fast path for trivial queries."* So the fast path exists, is tested, and is off by default — the feature was built and then not switched on.

## The fix

Default `brain.rules` to true. Anyone who wants every utterance to reach the model can still set it false explicitly.

## Acceptance criteria

The test must assert the **default configuration answers a clock query without calling the planner** — not merely that the field is true. A boolean check would pass any non-false value while proving nothing about routing, which is the same defect class as asserting a stage name without its duration.

Concretely: build a planner that fails the test if called, run `what time is it` through the assistant built from `config.Default()`, and require a spoken reply containing a time. A mutant restoring `Rules: false` must fail it.

## Measurement, not assertion

`stage=plan` now logs its duration, so the improvement is observable in the journal after deploy rather than claimed. Expected: the 7.1 s plan stage disappears for these queries.

jerryfane:
Closes #30.

## The measurement

The owner asked *"what time is it"* and waited about fourteen seconds:

| Stage | Measured |
|---|---|
| transcribe | 2.6–2.8 s |
| wake match | instant |
| **remote planner** | **7.102 s** |
| Piper synthesis | 0.908 s |
| post-agent, incl. playback | 3.825 s |

The Pi asked a language model across the internet for a value it holds in its own clock.

## The fast path already existed and shipped disabled

`internal/brain/rules.go` answers the time, the date, stop/cancel, and device commands with no model round-trip. `config.Default()` shipped:

```go
Rules:   false,   // "Rules enables the local no-model fast path for trivial queries."
```

Built, tested, and switched off. It defaults to true now; setting it false still sends everything to the planner.

## Two things the measurement exposed

**The phrasing was too narrow to be useful.** The clock matched `"what time"` or exactly `"time"`, so `"what is the time"` fell through to the model — the first version of this PR's own test caught that. A fast path that misses how people actually speak looks broken rather than absent. `asksClock`/`asksDate` now recognise the common forms, and the test asserts a timer request is **not** answered with the clock: `"set a timer for five minutes"` contains `time` and must reach the timer path.

**A real question still takes seconds, and silence during that wait is indistinguishable from Voice having missed the question.** `feedback.thinking` plays a quiet loopable cycle — `tick` (default) or `hum` — for the duration of the **planner call only**:

- not around the local fast path, which returns in milliseconds and would be heard as a glitch
- stopped before `Speak`, so the loop never talks over the answer
- rendered at startup and validated in config, so a typo is an error rather than silence discovered mid-answer

## The test I got wrong first

My loop test asserted only that no *new* plays started after `stop` — which a `stop` that merely cancels also satisfies. **The mutant passed.**

Real playback does not abandon a buffer the instant a context is cancelled, so the test now asserts nothing is **in flight** when `stop` returns:

```
--- FAIL: TestThinkingIsFinishedNotMerelyCancelledWhenStopReturns
    1 playback(s) still in flight when stop returned; the loop would overlap the reply
```

## Mutants, all failing

| Mutant | Fails |
|---|---|
| `Rules: false` restored | *the local fast path ships disabled…* |
| Clock phrasing narrowed to `"what time"` | *"what is the time" reached the planner* |
| `stop` cancels without waiting | *1 playback still in flight when stop returned* |

## Verification

`go test ./...` green; `CGO_ENABLED=1 go test -race` green on feedback, brain, audio, session; `go vet`, `gofmt -l` clean; exclusion-free `errcheck` exits 0.

## Expected effect, to be measured not claimed

`stage=plan` logs its duration, so after deploy the journal shows whether the 7.1 s stage disappears for clock and date queries. I will report the measured numbers rather than assert the improvement.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d65bdb5fa2e27c-1`

**Summary:** Reviewed jerryfane/voice PR #31 at exact head 7419f0446d275e2954c2476a238cc8133ba44352 (single commit, base f6747bb). Build/vet/gofmt/full test suite/errcheck all pass with CGO_ENABLED=0 (the module's intended pure-Go static build mode — no real cgo import exists; CGO_ENABLED=1 only fails here because this sandbox denies read access to /usr/include, an environment artifact unrelated to the PR). Rules-default-true, stop-then-Speak ordering, and the strengthened 'in-flight' loop test all check out. However, attack #1 (false local answers) is a real, confirmed regression: asksClock/asksDate use unanchored substring matching, so 'what time does the shop close', 'the time of the meeting', and 'what day should I book' are all answered locally with the current clock/date instead of reaching the planner — reproduced by execution, not just reading. This ships default-on (Brain.Rules=true), so it silently gives wrong answers to a plausible class of real questions. Attack #3 (thinking-loop lifecycle under cancellation) is fine: an added test simulating the whole session being cancelled mid-plan showed no goroutine leak and no double-close, and stop() correctly blocks until the in-flight buffer finishes. Attack #4 (test strength) — the corrected 'nothing in flight when stop returns' assertion is a genuine invariant of this implementation (stop blocks on a channel only closed after the loop goroutine's current PlayPCM call returns), not a probabilistic check; it is not too weak. Attack #2 (VAD/mic hallucination risk): static analysis of internal/vad/energy.go plus the shipped VAD defaults (MinSpeech=250ms, and any single non-speech frame resets the streak) shows neither thinking pattern's continuous-tone segment (tick ≈50ms, hum ≈180ms per ~1s cycle) can reach the continuity gate needed to become a segmented utterance sent to STT — the specific hallucination path is closed under defaults, though this is an unverified integration assumption (no test exercises concurrent playback+
[truncated]
**Severity:** `P1`

**Findings**
- **P1**
  - detail: asksClock matches unanchored substrings 'what time', 'the time', 'time is it'; asksDate matches 'what date', 'what day', 'the date', "what's the day", 'which day'. Reproduced with a throwaway test (TestReviewFalsePositiveLocalAnswers) run against a scratch copy of the repo at this exact head: 'what time does the shop close' -> Source=rules Speak='It is 3:04 PM.'; 'what day should I book' -> Source=rules Speak='It is Friday, September 18.'; 'the time of the meeting' -> Source=rules Speak='It is 3:04 PM.'. All three should have reached the planner and did not. Since config.Default().Brain.Rules is now true, this ships on by default.
  - evidence: executed
  - id: false-local-clock-date-answers
  - locator: internal/brain/rules.go:71-91 (asksClock, asksDate); reached via internal/brain/rules.go:25,28
- **P2**
  - detail: The comment at session.go:213-217 states the thinking sound 'runs only around the PLANNER, not around the local fast path,' but Thinking() is started before it is known whether Rules will answer locally. A throwaway test (TestThinkingCanFireOnAFastPathAnswer) driving a.accepted() with a rules-only planner that answers instantly showed PlayPCM fire in 1 of 500 trials despite zero model round-trip — the suppression is a timing race (win by cancelling before the goroutine's first PlayPCM call), not a structural guarantee. On real hardware, spawning and killing the playback subprocess mid-start could produce the audible artifact the feature was designed to avoid on exactly the class of request (fast local answers) it was meant to stay silent for.
  - evidence: executed
  - id: thinking-sound-not-scoped-to-planner-only
  - locator: internal/session/session.go:218-220 (stopThinking started before HandleText, which runs Rules.Plan first via internal/session/build.go: planner = brain.NewRules(ext))
- **P3**
  - detail: Under default config, neither thinking pattern's continuous-tone segment (tick ~50ms, hum ~180ms per ~1s cycle, both interrupted by long silence) can reach the 250ms uninterrupted-energy MinSpeech gate, and any single non-speech frame resets the streak — so under shipped defaults the tone cannot by itself produce a segmented utterance sent to STT. This clears the specific worst case raised, but it is a static/derived conclusion: no test in this PR or the existing suite exercises concurrent capture+playback, and real acoustic echo/room coupling on physical hardware (amplitude, reverberation tail) cannot be evaluated from this environment, so the conclusion holds only for the modeled waveform under default VAD settings.
  - evidence: executed
  - id: vad-mic-coupling-untested-assumption
  - locator: internal/vad/energy.go:100-138 (MinSpeech continuity gate, reset-on-non-speech-frame at line ~123); internal/config/config.go:268 (MinSpeech default 250ms); internal/audio/earcon.go:44,54-56 (tick/hum note timings)
- **P3**
  - detail: The 0.35 multiplier is hardcoded inside audio.Thinking() rather than surfaced through config, so a user raising feedback.volume to compensate for their room gets the thinking tone silently cut an additional 65% relative to what that same config field does for other sounds (earcon.go:94 applies volume unmodified). Worth documenting or exposing as its own ratio; not blocking.
  - evidence: static_only
  - id: thinking-volume-scaling-in-renderer
  - locator: internal/audio/earcon.go:44

**Tests Run**
- CGO_ENABLED=0 go build ./... at 7419f0446d275e2954c2476a238cc8133ba44352 — pass, no output
- gofmt -l . — clean, no files listed
- CGO_ENABLED=0 go test ./... — full suite green (all packages ok; internal/config, internal/faults, internal/proc have no test files by design)
- CGO_ENABLED=0 go run github.com/kisielk/errcheck@latest -blank -ignoretests -excludeonly ./... — exit 0, zero offences (independently confirms the exclusion-free-errcheck claim)
- go test -race ./internal/feedback/... ./internal/brain/... ./internal/audio/... ./internal/session/... — could NOT execute: cgo compilation failed (`fatal error: /usr/include/stdc-predef.h: Permission denied`) because this sandbox denies read access to system headers outside the workspace/tool-cache; this is an environment limitation, not a code defect, and is called out separately rather than claimed as executed
- go test ./internal/brain/... -run TestReviewFalsePositiveLocalAnswers -v (throwaway test added to a scratch copy of the repo at this exact head, not part of the PR) — FAILED as expected, reproducing 3/3 false-positive local answers for 'what time does the shop close', 'what day should I book', 'the time of the meeting'
- go test ./internal/feedback/... -run TestThinkingHandlesExternalContextCancellationMidPlan -v (throwaway test, scratch copy) — PASS: external context cancellation mid-plan produced no goroutine leak, no hang, no double-close on a subsequent stop() call
- go test ./internal/session/... -run TestThinkingCanFireOnAFastPathAnswer -v (throwaway test, scratch copy, 500 trials) — PASS but logged 'thinking sound fired on 1/500 fast-path-only trials', demonstrating the race described in the thinking-sound finding

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d65cff922f9411-1`

**Summary:** At head 9828ce0: f1 (false local clock/date answers) and f2 (thinking-sound race) are genuinely fixed and executed-verified, including a mutant-kill test for f2 that reproduced the requester's exact claimed failure mode. f4 is resolved by documentation as scoped. f3 is unchanged and still open (P3, non-blocking). However, the new content-word rule in rules.go tokenizes on strings.Fields() without stripping punctuation, and the only normalization applied to STT transcripts anywhere in the pipeline is ToLower+TrimSpace (rules.go:23; session.go:56) — so realistic whisper.cpp output with terminal punctuation ('What time is it?', 'Time.') or apostrophes ('What's the time?') misses the local fast path and falls back to the remote planner. This is a regression versus the prior substring-based implementation, which was immune to trailing punctuation. It fails safe (no wrong answer) but plausibly defeats the PR's core purpose for a large share of real transcripts.
**Severity:** `P2`

**Findings**
- **False local clock/date answers on unrelated questions** (P1)
  - continues_uid: jerryfane/voice#31-f1
  - detail: asksClock/asksDate at rules.go:104-112 now gate on content(n) having exactly one non-filler word equal to 'time'/'date'/'day', eliminating the unanchored substring matches that previously fired on 'what time does the shop close' etc.
  - file: internal/brain/rules.go
  - state: withdrawn
  - withdraw_reason: Fixed: asksClock/asksDate (internal/brain/rules.go:104-112) now require content(n) to have exactly one word equal to the query subject, replacing substring Contains matching. Verified by running TestShippedDefaultAnswersTheClockWithoutAModel (internal/brain/rules_test.go), which asserts the exact three original repro strings ('what time does the shop close', 'the time of the meeting', 'what day should I book') plus three more variants now reach the planner instead of answering locally — test passes (CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v).
- **Thinking sound scoped to planner only, no longer a race** (P2)
  - continues_uid: jerryfane/voice#31-f2
  - detail: The 400ms delay gates the playback goroutine before it ever calls PlayPCM, and mutating it out reproduces the originally-reported race failure exactly, confirming both the fix and the falsifiability of the new test.
  - file: internal/feedback/feedback.go
  - state: withdrawn
  - withdraw_reason: Fixed structurally: internal/feedback/feedback.go:106-125 now gates the entire play loop behind select{ <-loop.Done(); <-time.After(thinkingDelay) }, so PlayPCM cannot be called before 400ms elapses — not a race to cancel. Verified by running TestThinkingStaysSilentForAFastAnswer and TestThinkingPlaysForASlowAnswer (both pass), then applying the exact 'no-delay mutant' (removed the select block) in a scratch copy and rerunning TestThinkingStaysSilentForAFastAnswer: it failed with 7 plays in 100ms and 7 plays across 200 immediate calls, matching the requester's own claimed mutant-kill count exactly. Restored the original file and reran all feedback Thinking tests — pass.
- **VAD/mic coupling assumption remains untested on hardware** (P3)
  - continues_uid: jerryfane/voice#31-f3
  - detail: No file in this diff (internal/audio/earcon.go, internal/brain/rules.go, internal/brain/rules_test.go, internal/config/config.go, internal/feedback/feedback.go, internal/feedback/feedback_test.go) touches internal/vad/energy.go, and config.go's MinSpeech default is confirmed unchanged at 250ms (config.go:273). The static/derived conclusion from the prior round stands unmodified; still requires on-device acoustic verification, which the requester has already acknowledged they will do post-deploy.
  - file: internal/vad/energy.go
  - state: answered
- **Thinking volume ratio hardcoded but now documented** (P3)
  - continues_uid: jerryfane/voice#31-f4
  - detail: earcon.go:34 and config.go's Feedback.Thinking doc now name and explain the 0.35 ratio; the value itself remains a deliberate, non-configurable design choice as stated in the code comment.
  - file: internal/audio/earcon.go
  - state: withdrawn
  - withdraw_reason: The original ask was documentation/visibility ('worth documenting or exposing as its own ratio; not blocking'), not a change to the hardcoded value. internal/audio/earcon.go:34 now names the constant thinkingVolumeRatio = 0.35 with an explanatory comment, and internal/config/config.go's Feedback.Thinking doc comment explicitly states the ratio and that raising Volume raises both proportionally. The documentation gap is closed.
- **Content-word rule has no punctuation/apostrophe normalization, so realistic STT transcripts silently miss the local fast path** (P2)
  - detail: asksClock/asksDate tokenize via strings.Fields() over text normalized only by ToLower+TrimSpace (rules.go:23,87-95; upstream session.go:56 also only TrimSpaces). Punctuation attached to a token is never stripped, so it becomes part of the token and fails both the filler-map lookup and the single-content-word check. Executed against a scratch copy of the repo at this exact head (9828ce0) via a throwaway test (TestReviewProbePunctuationFromRealSTT) run with `CGO_ENABLED=0 go test ./internal/brain/... -run TestReviewProbePunctuationFromRealSTT -v`: 'What time is it?' -> Source!=rules (reached planner), 'What's the time?' -> Source!=rules, 'Time.' -> Source!=rules. All three should answer locally per the PR's own stated intent and did not. config.Default().STT.Command (config.go:287-294) invokes whisper-cli with no punctuation-suppressing flag; whisper/whisper.cpp is documented to emit terminal punctuation (periods, question marks) by default, making these not adversarial inputs but plausible everyday transcripts. This is a regression introduced by this diff: the prior substring-based asksClock/asksDate (Contains(n, "what time") etc.) was immune to trailing punctuation because it does not require an exact token match. The defect fails safe (falls back to the correct but slower remote planner, no wrong answer), but it plausibly undermines the PR's stated purpose — eliminating the model round-trip for trivial queries — for a large share of real spoken utterances.
  - file: internal/brain/rules.go

**Tests Run**
- CGO_ENABLED=0 go vet ./... (on reviewed worktree at 9828ce0) — clean, exit 0
- CGO_ENABLED=0 gofmt -l . (on reviewed worktree) — no files listed, clean
- CGO_ENABLED=0 go build ./... (on reviewed worktree) — exit 0
- CGO_ENABLED=0 go test ./... (on reviewed worktree) — all packages ok, including internal/brain and internal/feedback
- CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v (on reviewed worktree) — PASS, confirms f1's repro strings plus new variants now reach the planner
- git archive 9828ce0 into a writable scratch dir (worktree itself is write-protected even with sandbox disabled), added throwaway internal/brain/review_probe_test.go and review_probe2_test.go, ran CGO_ENABLED=0 go test ./internal/brain/... -run TestReviewProbeContentRule -v — revealed "what's the time", "what's the day", and "whats todays date" fail to reach the local path
- CGO_ENABLED=0 go test ./internal/brain/... -run TestReviewProbePunctuationFromRealSTT -v (scratch copy) — confirmed 'What time is it?', 'What's the time?', 'Time.' all incorrectly reach the planner instead of answering locally
- CGO_ENABLED=0 go test ./internal/feedback/... -run TestThinking -v (scratch copy) — all 4 tests pass, including the new f2 tests
- Applied a no-delay mutant to feedback.go (removed the select{loop.Done/time.After} gate) in the scratch copy, reran TestThinkingStaysSilentForAFastAnswer — FAILED with 7 plays in 100ms and 7 plays across 200 immediate calls, matching the requester's claimed mutant behavior exactly; restored the original file and reconfirmed all tests pass
- go install github.com/kisielk/errcheck@latest into a scratch GOBIN, ran CGO_ENABLED=0 errcheck -blank -ignoretests -excludeonly ./... against the reviewed worktree at 9828ce0 — zero output, exit 0, confirming the exclusion-free errcheck claim
- Attempted CGO_ENABLED=1 go test -race ./internal/feedback/... ./internal/brain/... — could not execute: this sandbox's gcc cannot read /usr/include/stdc-predef.h (environment limitation unrelated to the PR), so the requester's 'race green' claim could not be independently reproduced and is taken on their word
- git status on the reviewed worktree before and after all testing — confirmed clean throughout; no files in the reviewed worktree were modified

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d65dc5ab09d901-1`

**Summary:** PR #31 at head 9a8e0c8, diffed against prior-seen head 9828ce0 (internal/brain/rules.go + rules_test.go only). f5's punctuation/apostrophe normalisation gap is genuinely fixed and both requester-claimed mutants were independently reproduced and killed. However, widening the same normalise() call to govern ALL of Plan()'s routing (not just clock/date) introduced a new regression: device-ID substring matching now silently breaks for any device ID containing punctuation, because the transcript is normalised but device.Info.ID is not. It fails safe (falls to the planner with the original text) but defeats the fast path for such devices — the same failure shape this PR exists to fix, reintroduced one layer over. Full build/vet/gofmt/test/errcheck(CI-pinned v1.9.0)/static-binary checks all pass; go test -race could not run due to a sandbox restriction on cgo (unrelated to the code). Same-model-family caveat applies: this review was performed by a Claude model reviewing Claude-authored code.
**Severity:** `P2`

**Findings**
- **False local clock/date answers on unrelated questions** (P1)
  - continues_uid: jerryfane/voice#31-f1
  - detail: content()/asksClock()/asksDate() (rules.go:117-141) are byte-for-byte unchanged by this diff (git diff 9828ce0..9a8e0c8 touches only the normalise() addition and the `n :=` line). Reran `CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v` at 9a8e0c8: PASS, including all seven false-positive repro strings from the original report. Fix stands.
  - file: internal/brain/rules.go
  - state: answered
- **Thinking sound scoped to planner only, no longer a race** (P2)
  - continues_uid: jerryfane/voice#31-f2
  - detail: internal/feedback/feedback.go has zero diff lines between 9828ce0 and 9a8e0c8 (verified via `git diff 9828ce0..9a8e0c8 -- internal/feedback/feedback.go | wc -l` = 0). Ran full suite `CGO_ENABLED=0 go test ./...`: internal/feedback ok (0.977s), including the Thinking tests. No change, fix stands.
  - file: internal/feedback/feedback.go
  - state: answered
- **VAD/mic coupling assumption remains untested on hardware** (P3)
  - continues_uid: jerryfane/voice#31-f3
  - detail: internal/vad/energy.go and internal/config/config.go both have zero diff lines between 9828ce0 and 9a8e0c8; the only files touched by this diff are internal/brain/rules.go and rules_test.go (confirmed via `git diff 9828ce0..9a8e0c8 --stat`). Static conclusion unchanged; still requires on-device acoustic verification the requester has acknowledged doing post-deploy.
  - file: internal/vad/energy.go
  - state: answered
- **Thinking volume ratio hardcoded but now documented** (P3)
  - continues_uid: jerryfane/voice#31-f4
  - detail: internal/audio/earcon.go has zero diff lines between 9828ce0 and 9a8e0c8. The documented thinkingVolumeRatio = 0.35 constant is unchanged this round; documentation gap remains closed.
  - file: internal/audio/earcon.go
  - state: answered
- **Content-word rule has no punctuation/apostrophe normalization, so realistic STT transcripts silently miss the local fast path** (P2)
  - continues_uid: jerryfane/voice#31-f5
  - detail: See withdraw_reason.
  - state: withdrawn
  - withdraw_reason: Fixed by this exact diff (9828ce0->9a8e0c8). Verified three ways at head 9a8e0c8 in a disposable full-repo scratch copy: (1) `CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v` passes, including the exact three original repro strings 'What time is it?', 'What's the time?', 'Time.' plus 'What time does the shop close?' now added to the false-positive list. (2) Mutant reverting `n := normalise(text)` to the pre-fix `strings.ToLower(strings.TrimSpace(text))`: test fails with 'What time is it?' reaching the planner, exactly as the requester claimed. (3) Mutant removing the apostrophe-deletion case (letting apostrophes fall through to the default space-replacement branch): test fails with 'What's the time?' reaching the planner, exactly as the requester claimed. Also independently confirmed errcheck (CI-pinned `go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./...`) exits 0, validating the rune-slice-instead-of-strings.Builder claim.
- **normalise() now governs all of Plan()'s routing, silently breaking device-ID matching for IDs containing punctuation** (P2)
  - detail: rules.go:24 changed `n := strings.ToLower(strings.TrimSpace(text))` to `n := normalise(text)`, and `n` is the single variable used for EVERY branch of Plan() — clock/date, stop/cancel, AND the device-ID/action-keyword loop at rules.go:33-59. normalise() (rules.go:74-90) strips all punctuation (turning e.g. "kid's room" into "kids room" and "front-door" into "front door"), but the device ID it's matched against — `id := strings.ToLower(d.ID)` at rules.go:36 — is never run through normalise() and keeps its punctuation. So `strings.Contains(n, id)` at rules.go:37 now fails for any device ID containing punctuation that appears verbatim in the transcript. Proved as a REGRESSION (not pre-existing) by direct before/after test in a scratch copy: with device.Info{ID: "kid's room", Capabilities: ["on","off"]} and input "turn on the kid's room", checking out rules.go/rules_test.go at prior head 9828ce0 into the scratch tree gives Source="rules", Speak="Okay, on kid's room." (matches, correct); rebuilding at current head 9a8e0c8 with the identical fixture and input gives Source="", Speak="from the model" (falls through to the planner). Same pattern reproduced for a hyphenated ID ("front-door"). Fails safe — Plan()'s final fallback `return r.Next.Plan(ctx, text, inv)` passes the ORIGINAL un-normalised text to the downstream planner, so no wrong action results, only the lost fast path — but it is the identical failure class this PR exists to fix, reintroduced one call deeper. Blast radius is narrower than f5's (only devices whose configured ID contains punctuation, e.g. apostrophe- or hyphen-named rooms), but the regression is unambiguous and verifiable.
  - file: internal/brain/rules.go
- **Three independently-drifting normalise/normalize implementations (brain, wake, timer)** (P3)
  - detail: This diff adds a third distinct normaliser: internal/wake/match.go:127 (normalize, ASCII lower via unicode.ToLower, keeps letter/digit/space via unicode classes, turns apostrophes into SPACES not deletion), internal/timer/parse.go:311 (normalise, ASCII-only a-z/0-9 via byte comparison, also turns apostrophes into spaces), and now internal/brain/rules.go:74 (normalise, unicode.IsLetter/IsDigit so non-Latin scripts are kept as letters, and uniquely DELETES apostrophes rather than spacing them). Confirmed via `git diff 9828ce0..9a8e0c8 -- internal/timer/parse.go internal/wake/match.go` = 0 lines that neither of the other two was touched by this diff, so this is not a new-diff defect, but the diff does deepen an existing drift: three packages now each reimplement 'strip punctuation, lowercase, collapse whitespace' with three different apostrophe rules and two different letter-class scopes (unicode vs ASCII-only), which is the same shape of drift that caused the ExactOnly/Match divergence bug fixed in #23. Not a live bug today since each consumer's matching logic (exact single-content-word vs sound-alike skeleton vs numeric-duration parsing) tolerates its own normaliser's quirks, and a naive merge could silently change behavior (e.g. timer's ASCII-only pass would reject non-Latin duration words wake/brain now accept). Worth a deliberate shared-utility follow-up rather than blocking this PR.
  - file: internal/brain/rules.go

**Tests Run**
- git diff 9828ce0..9a8e0c8 --stat -- (confirms only internal/brain/rules.go + rules_test.go changed)
- git diff 9828ce0..9a8e0c8 -- internal/timer/parse.go internal/wake/match.go (0 lines, confirms untouched)
- git diff 9828ce0..9a8e0c8 -- internal/feedback/feedback.go internal/vad/energy.go internal/config/config.go internal/audio/earcon.go (0 lines each, confirms f2/f3/f4 unaffected)
- CGO_ENABLED=0 go build ./... -> success
- CGO_ENABLED=0 go vet ./... -> clean, exit 0
- gofmt -l . -> no output, clean
- CGO_ENABLED=0 go test ./internal/brain/... -v -> PASS (TestShippedDefaultAnswersTheClockWithoutAModel)
- CGO_ENABLED=0 go test ./... -> full suite green across all 13 packages with tests
- go test -race ./internal/brain/... -> build failed: sandbox denies read access to /usr/include/stdc-predef.h for cgo; environment limitation, not code defect, could not execute
- go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... (exact CI invocation from .github/workflows/ci.yml) -> exit 0, clean
- CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ... ./cmd/voice -> success
- CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o ... ./cmd/voice -> success
- Scratch copy (full git checkout at 9a8e0c8 in a writable side directory) + throwaway TestReviewProbeNormalise: exercised o'clock/curly-apostrophe, trailing 'thank you', leading 'um', hyphenation ('twenty-three'), em-dash, ellipsis, guillemets, curly quotes, French, Japanese, digits, accented Latin, Cyrillic -> all fail-safe, no false positives found
- Mutant 1 in scratch copy: reverted `n := normalise(text)` to `strings.ToLower(strings.TrimSpace(text))` -> TestShippedDefaultAnswersTheClockWithoutAModel FAILS with 'What time is it?' reaching the planner, exactly as claimed
- Mutant 2 in scratch copy: removed the apostrophe-deletion case in normalise() -> TestShippedDefaultAnswersTheClockWithoutAModel FAILS with 'What's the time?' reaching the planner, exactly as claimed
- Scratch-copy regression test TestReviewProbeDeviceIDPunctuation with device.Info{ID:"kid's room"}: at head 9a8e0c8 -> Source="" (falls to planner, FAIL); after checking out rules.go/rules_test.go at prior head 9828ce0 into the same scratch tree with an identical test -> Source="rules" (PASS), proving the regression is introduced by this diff
- Scratch-copy TestReviewProbeHyphenDeviceID with device.Info{ID:"front-door"} at head 9a8e0c8 -> Source="" (falls to planner), same pattern confirmed for hyphenated IDs
- Post-check: `git status --short` on the actual reviewed workspace -> clean, HEAD still 9a8e0c8a2b921c99199a97113ae5571f25f5d7a7, confirming no files were edited in the reviewed repo

jerryfane:
## `#31-f6` fixed, `#31-f7` deferred to issue #32, at head `07af195e1bd97a86bdce3f230fa740ddef6f0aac`

### `#31-f6` (P2) — a regression I introduced one round earlier

Correct and important. Widening `normalise` to govern all of `Plan` meant the **transcript** was normalised while the **device ID** was only lowercased, so every device whose ID contains punctuation stopped matching: `"turn on the living-room lamp"` normalises to `living room`, compared against `living-room`, matched nothing.

The fast path would have silently stopped controlling those devices while answering the clock faster — a trade nobody asked for and nobody would have noticed until a light stopped responding.

Normalising **one side** is how a shared rule becomes a silent regression. Both sides go through `normalise` now, and an empty ID is skipped rather than matching everything.

The test uses IDs that actually occur and asserts both spoken and written forms reach the right device:

```
MUTANT comparing a raw ID against normalised text:
  "turn on the living-room" produced no action; the device was not recognised
  "turn on living room"     produced no action; the device was not recognised
  "turn on lamp_2"          produced no action; the device was not recognised
```

### `#31-f7` (P3) — deferred to issue #32, with the reason

You are right that three normalisers is the `#23-f9` shape: two implementations of one concept stay equal only until someone fixes a bug in one. They already disagree — `brain` deletes apostrophes so `"what's"` reduces to one filler word, `wake` spaces them because nothing there depends on it yet.

Deferred because consolidating them touches three packages and **changes `wake`'s merged, independently approved behaviour**. `wake` was hardened over six rounds and two collisions no threshold could fix; altering its normalisation as the fourth round of a latency PR is how that gets undone quietly. Meanwhile the owner's device pays 7 seconds for a clock query while this waits.

Issue #32 carries the fix and an acceptance criterion that matters more than the consolidation: **a test proving the wake gate and the planner see the same normalised text for one transcript**, so drift is caught rather than assumed absent — plus `wake`'s existing tests passing unchanged, or each behaviour change named rather than absorbed.

## Verification at this head

`go test ./...` green; `go vet`, `gofmt -l` clean; exclusion-free `errcheck` exits 0; CI `test` SUCCESS.

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d65ebcf20fd64b-1`

**Summary:** PR31 head 07af195, diffed against prior-seen 9a8e0c8: the round's only change (rules.go id-normalisation fix) genuinely fixes f6 as claimed, verified with an overlay-based mutant. But normalising both sides introduces a new, live collision: two configured device IDs that differ only in punctuation style (e.g. 'lamp-2' vs 'lamp_2') now normalise to the identical string, and the registry (device.go Add(), keyed on lowercased-but-not-normalised ID) does not prevent both from coexisting. Reproduced against the real Plan() via a go build overlay (no repo files modified, git status stayed clean throughout): with both IDs in inventory, 'turn on lamp 2' actuated whichever device happened to be first in the slice - an arbitrary wrong-device actuation, which is a worse failure mode than the miss this round fixed, even though its blast radius (requires two IDs to literally collide post-normalisation) is much narrower than f6's. Full build/vet/gofmt/test suite and errcheck all pass. f7's deferral-to-#32 reasoning is sound and unchanged by this round. Same-model-family caveat: this review is a Claude model reviewing Claude-authored code.
**Severity:** `P2`

**Findings**
- **False local clock/date answers on unrelated questions** (P1)
  - continues_uid: jerryfane/voice#31-f1
  - detail: content()/asksClock()/asksDate() are outside this round's diff: `git diff 9a8e0c8..07af195 --stat` shows only rules.go (loop lines 33-59 + doc comment) and rules_test.go changed. Ran `CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v` at 07af195: PASS. Fix stands, unaffected by this round.
  - file: internal/brain/rules.go
  - state: answered
- **normalise() now governs all of Plan()'s routing, silently breaking device-ID matching for IDs containing punctuation** (P2)
  - continues_uid: jerryfane/voice#31-f6
  - state: withdrawn
  - withdraw_reason: Fixed by this exact diff (9a8e0c8->07af195). rules.go:41 now reads `id := normalise(d.ID)` (was `strings.ToLower(d.ID)`), with an `id == ""` skip guard. Verified three ways: (1) `CGO_ENABLED=0 go test ./internal/brain/... -run TestDeviceIDsWithPunctuationStillMatch -v` passes for all five punctuated/spoken-form pairs (living-room, lamp_2, hue:desk). (2) Built a go-build-overlay mutant reverting rules.go:41 to `strings.ToLower(d.ID)` (no repo files touched, `git status --porcelain` stayed empty) and reran the same test: all five cases fail with 'produced no action; the device was not recognised', reproducing exactly the regression f6 described. (3) `CGO_ENABLED=0 go vet ./...`, `gofmt -l .`, `CGO_ENABLED=0 go build ./...`, full `CGO_ENABLED=0 go test ./...`, and `go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./...` are all clean at this head.
- **Normalising the device ID as well as the transcript creates a new ID-collision class: two configured IDs that differ only by punctuation style now match identically and can pick the wrong device** (P2)
  - detail: normalise() strips all punctuation to spaces (except deleted apostrophes), so normalise("lamp-2") == normalise("lamp_2") == "lamp 2", and normalise("hue:desk") == normalise("hue desk"). Confirmed with a standalone extraction of normalise() in a scratch Go program. Then reproduced against the real Plan() with a go build overlay adding a probe test file at internal/brain (repo left untouched - `git status --porcelain` empty before and after): with inv = [{ID:"lamp-2"},{ID:"lamp_2"}] and input "turn on lamp 2", Plan() returned Actions[0].Device == "lamp-2" - the first inventory entry - not a considered choice, purely slice order. Nothing in internal/device/device.go's Registry.Add() (keyed on `strings.ToLower(dev.ID())`, not normalise()) prevents two such colliding devices from being registered simultaneously, so this is reachable with a real config, not just a contrived test fixture. Distinct from f6/f5: those failed safe (fell through to the downstream planner, no wrong action); this one actuates an actually-wrong physical device, which is a worse failure category even though it needs a narrower precondition (two device IDs whose punctuation-stripped forms collide) to trigger. Separately confirmed the short-ID-substring-inside-longer-ID behavior the requester also asked about (e.g. "lamp" matching inside normalised "lamp 2") is unchanged pre-existing behavior from before 9a8e0c8 (id comparison for unpunctuated IDs was already `strings.Contains(n, lowercased_id)`), so that part is not a new defect introduced by this diff and is not filed separately. Checked the rest of Plan() for the same raw-vs-normalised defect class the requester asked about (item 3): the clock/date/stop checks compare `n` only against literal, already-punctuation-free filler words, and actionPlan()'s use of raw `d.ID` (not normalised) for Action.Device/Speak is correct and necessary, since Registry.Get() resolves by lowercased-but-not-normalised ID - normalising there would break downstrea
[truncated]
  - file: internal/brain/rules.go
- **Three independently-drifting normalise/normalize implementations (brain, wake, timer)** (P3)
  - continues_uid: jerryfane/voice#31-f7
  - detail: Confirmed via `git diff 9a8e0c8..07af195 -- internal/wake/match.go internal/timer/parse.go internal/vad/energy.go internal/config/config.go internal/audio/earcon.go internal/feedback/feedback.go | wc -l` = 0 that none of these files changed this round. The deferral to issue #32 is sound: consolidating now would touch wake's independently-hardened apostrophe-to-space behavior as a side effect of a latency PR, and the stated acceptance criterion (a same-transcript proof test across wake and brain, plus wake's existing tests passing unchanged or each behavioral change named rather than silently absorbed) is the right bar. No pushback on that reasoning. Still open, still non-blocking per policy.
  - file: internal/brain/rules.go
  - state: answered
- **Thinking sound scoped to planner only, no longer a race** (P2)
  - continues_uid: jerryfane/voice#31-f2
  - detail: Zero diff in internal/feedback/feedback.go since 9a8e0c8 (confirmed above). Full suite run this round: `CGO_ENABLED=0 go test ./...` shows internal/feedback ok. Unaffected by this round, fix stands.
  - file: internal/feedback/feedback.go
  - state: answered
- **VAD/mic coupling assumption remains untested on hardware** (P3)
  - continues_uid: jerryfane/voice#31-f3
  - detail: Zero diff in internal/vad/energy.go and internal/config/config.go since 9a8e0c8 (confirmed above). Static conclusion unchanged; still requires on-device acoustic verification, unaffected by this round's rules.go-only diff.
  - file: internal/vad/energy.go
  - state: answered
- **Thinking volume ratio hardcoded but now documented** (P3)
  - continues_uid: jerryfane/voice#31-f4
  - detail: Zero diff in internal/audio/earcon.go since 9a8e0c8 (confirmed above). thinkingVolumeRatio = 0.35 constant and its documentation are unchanged this round.
  - file: internal/audio/earcon.go
  - state: answered

**Tests Run**
- git diff 9a8e0c8..07af195 --stat (confirms only internal/brain/rules.go and rules_test.go changed)
- git diff 9a8e0c8..07af195 -- internal/wake/match.go internal/timer/parse.go internal/vad/energy.go internal/config/config.go internal/audio/earcon.go internal/feedback/feedback.go | wc -l => 0
- CGO_ENABLED=0 go build ./... => clean
- CGO_ENABLED=0 go vet ./... => clean
- gofmt -l . => no output
- CGO_ENABLED=0 go test ./... => all packages ok
- CGO_ENABLED=0 go test ./internal/brain/... -run TestDeviceIDsWithPunctuationStillMatch -v => PASS
- CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v => PASS
- go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... => exit 0
- go build overlay mutant: rules.go id := normalise(d.ID) reverted to strings.ToLower(d.ID), rerun TestDeviceIDsWithPunctuationStillMatch => FAIL on all 5 cases (proves fix necessity; repo files never modified, git status --porcelain empty throughout)
- standalone scratch Go program reproducing normalise('lamp-2') == normalise('lamp_2') == 'lamp 2' and normalise('hue:desk') == normalise('hue desk')
- go build overlay probe test against real Plan(): inv=[{ID:'lamp-2'},{ID:'lamp_2'}], input 'turn on lamp 2' => Actions[0].Device=='lamp-2' (arbitrary first-match, proves new collision; repo untouched)
- go build overlay probe test: inv=[{ID:'lamp'},{ID:'lamp_2'}], input 'turn on lamp 2' => Actions[0].Device=='lamp' (confirms pre-existing substring behavior, unrelated to this diff)
- git status --porcelain checked clean before/after every overlay-based probe to confirm no repo files were edited

jerryfane:
## `#31-f8` answered at head `987ebe072b348c665a004b7d20344230a0f275b0`

This is the collision I asked you to hunt after F6, and you found it live: normalising both sides made `lamp-2` and `lamp_2` reduce to the same text, and the first match in inventory order won silently. **Acting on the wrong light is worse than the miss normalising fixed**, because the user gets a confident "Okay" while the wrong thing happens.

Matching devices are now collected rather than acted on as found:

- one match acts exactly as before
- more than one **asks**, naming both IDs so the question can actually be answered, and takes no action at all

The test asserts no action *and* that both IDs appear in the question — a clarification that does not name the options is not a clarification — *and* that a single match still acts, so the guard cannot quietly become a wall.

## On the mutant, because my first one was invalid

My first attempt at the mutant deleted a closing brace and did not compile:

```
internal/brain/rules.go:57:2: syntax error: non-declaration statement outside function body
```

A non-compiling mutant proves nothing, and I nearly recorded it as a kill. The valid mutant weakens the guard to `len(named) > 99` so first-match-wins is restored, and it fails:

```
--- FAIL: TestAnAmbiguousDeviceNameAsksInsteadOfGuessing
    acted on "lamp-2" despite two devices matching; it cannot know which
```

## Still open from you, and I have not addressed it in this round

You raised that `Contains` means a short normalised ID matches inside a longer one — `lamp` inside `lamp 2`. That is a **pre-existing** property of this matcher rather than something normalising introduced, and it is not what F8 named. If you think it belongs in this PR say so; otherwise I would file it separately rather than widen a latency change a fifth time, and I would rather hear your call than assume mine.

## Verification at this head

`go test ./...` green; `CGO_ENABLED=1 go test -race ./internal/brain/` green; `go vet`, `gofmt -l` clean; exclusion-free `errcheck` exits 0.

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d65f6c8fdec0a4-1`

**Decision:** `approved`
**Summary:** PR #31 at exact head 987ebe072b348c665a004b7d20344230a0f275b0 (round 5, diff from 07af195e1bd9 confined to internal/brain/rules.go + rules_test.go). f8's fix verified: matching devices are now collected before acting; >1 match asks naming every matching ID and takes no action, a single match still acts. Confirmed via the shipped unit test, an independently-built valid mutant (weakening the guard to len(named)>99) that compiles cleanly and is killed reproducing the original bug, and a clean full suite/vet/gofmt/errcheck. Answered all three of the requester's attack angles: (1) the lamp/lamp-2 substring case does force an ask, but overlay-probing the SAME input against the prior head shows it previously silently acted on the wrong device (lamp), so this is an improvement, not a regression - filed as a new non-blocking P3 refinement opportunity (prefer longest/exact match before asking); (2) asking directly rather than falling through to the external planner is defensible since Rules must work with no model available at all, and there's no evidence the planner would disambiguate better; (3) the question does speak raw punctuated IDs verbatim through an arbitrary external TTS binary with no normalisation anywhere in the path - unverifiable in this sandbox (no TTS/audio access), filed as informational P3, not confirmed broken and not a new pattern (single-match confirmations already did this pre-PR). f1 (P1) and f7 (P3) re-checked and unaffected by this round's diff; f2/f3/f4 likewise zero-diff and unaffected. Could not execute go test -race in this sandbox (cgo blocked environment-wide: 'stdc-predef.h: Permission denied', confirmed with a raw gcc invocation unrelated to this code) so the requester's 'race green' claim is unverified by me, not contradicted. Same-model-family caveat: this review was performed by a Claude runtime; the PR's commit history is consistent with Claude authorship.

**Findings**
- **Normalising the device ID as well as the transcript creates a new ID-collision class** (P2)
  - continues_uid: jerryfane/voice#31-f8
  - file: internal/brain/rules.go
  - state: withdrawn
  - withdraw_reason: Fixed by this exact diff (07af195->987ebe0). rules.go now collects all ID-matching devices into `named` (lines 40-47) before acting; len(named)>1 asks naming every matching raw ID and takes no action (lines 48-57); a single match still acts. Verified: (1) `CGO_ENABLED=0 go test ./internal/brain/... -run TestAnAmbiguousDeviceNameAsksInsteadOfGuessing -v` passes: zero actions on ambiguity, both IDs present in the question, single match still acts. (2) Built a go-build-overlay mutant weakening rules.go:48 from `len(named) > 1` to `len(named) > 99` (no repo files touched, `git status --porcelain` empty throughout) - it compiles cleanly (a VALID mutant, unlike the requester's first non-compiling attempt) and the same test fails with 'acted on "lamp-2" despite two devices matching', exactly reproducing the collision f8 named. (3) `CGO_ENABLED=0 go build ./...`, `go vet ./...`, `gofmt -l .`, full `go test ./...` (all packages ok), and `go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./...` are all clean at this head.
- **Ambiguity guard treats a strict-substring ID ("lamp") as equally ambiguous with a longer exact match ("lamp-2"), forcing a needless ask - but confirmed not a regression versus the prior head** (P3)
  - detail: Reproduced with a go-build-overlay probe test (no repo files touched) at inv=[{ID:"lamp"},{ID:"lamp-2"}], input "turn on lamp 2": at this head Plan() returns Speak="I know more than one device by that name: lamp, lamp-2. Which one?", Actions=[] - both match via normalise+Contains (rules.go:42-43) purely because "lamp" is a substring of "lamp 2". Ran the identical probe against the prior head's rules.go (07af195, via overlay) with the same inventory: it silently returned Speak="Okay, on lamp.", Actions=[{Device:lamp,Op:on}] - the WRONG device, no ask, first-match-wins. So the requester's attack #1 concern is real (the guard asks on a case a person would consider unambiguous) but is NOT a regression: the prior behaviour for this exact input was a silent wrong actuation, which is strictly worse. The underlying Contains-based substring matching predates this PR (noted pre-existing in an earlier review round) and the guard has simply made that latent imprecision visible as a spoken question instead of a silent wrong action. Preferring the longest/most-specific ID match before falling back to asking is a reasonable follow-up, out of scope for a guard whose job was to stop confident wrong actuation, which it does.
  - file: internal/brain/rules.go
- **The disambiguation question speaks raw device IDs verbatim, including punctuation, through an arbitrary external TTS command with no normalisation anywhere in the path** (P3)
  - detail: rules.go:50-54 builds the question from `d.ID` unmodified (e.g. "hue:desk", "lamp_2"). internal/session/session.go:78-86 (Speak) passes Plan.Speak verbatim to Synthesizer.Synthesize, implemented in internal/speech/command.go as an arbitrary external CLI (CommandSynthesizer) - no speech normalisation exists anywhere on this path. Whether "lamp_2" comes out as a sane spoken form or an awkward literal reading depends entirely on the configured TTS engine, unverifiable in this sandbox (no TTS binary, no audio hardware - same category as the existing f3 on-device-only caveat). Not a confirmed defect either way; the same raw-ID-in-Speak pattern already existed for single-match confirmations ("Okay, on lamp-2.") before this PR, so this is not a new departure from existing behaviour.
  - file: internal/brain/rules.go
- **False local clock/date answers on unrelated questions** (P1)
  - continues_uid: jerryfane/voice#31-f1
  - detail: content()/asksClock()/asksDate() (rules.go:135-161) are untouched by this round's diff (07af195->987ebe0 touches only the device-matching loop, lines ~32-59). Re-ran `CGO_ENABLED=0 go test ./internal/brain/... -run TestShippedDefaultAnswersTheClockWithoutAModel -v` at 987ebe0: PASS. Fix stands, unaffected by this round.
  - file: internal/brain/rules.go
  - state: answered
- **Three independently-drifting normalise/normalize implementations (brain, wake, timer)** (P3)
  - continues_uid: jerryfane/voice#31-f7
  - detail: `git diff 07af195..987ebe0 -- internal/wake/match.go internal/timer/parse.go internal/vad/energy.go internal/config/config.go internal/audio/earcon.go internal/feedback/feedback.go` = 0 lines: no sibling normalise implementation or unrelated module changed this round. Deferral to issue #32 remains sound and non-blocking; still open, unaffected by this round.
  - file: internal/brain/rules.go
  - state: answered
- **Thinking sound scoped to planner only, no longer a race** (P2)
  - continues_uid: jerryfane/voice#31-f2
  - detail: Zero diff in internal/feedback/feedback.go since 07af195 (confirmed via the same git diff --stat check). Full suite this round: `CGO_ENABLED=0 go test ./...` shows internal/feedback ok. Unaffected by this round, fix stands.
  - file: internal/feedback/feedback.go
  - state: answered
- **VAD/mic coupling assumption remains untested on hardware** (P3)
  - continues_uid: jerryfane/voice#31-f3
  - detail: Zero diff in internal/vad/energy.go and internal/config/config.go since 07af195. Static conclusion unchanged; still requires on-device acoustic verification, unaffected by this round's rules.go-only diff.
  - file: internal/vad/energy.go
  - state: answered
- **Thinking volume ratio hardcoded but now documented** (P3)
  - continues_uid: jerryfane/voice#31-f4
  - detail: Zero diff in internal/audio/earcon.go since 07af195. thinkingVolumeRatio = 0.35 constant and its documentation are unchanged this round.
  - file: internal/audio/earcon.go
  - state: answered

**Tests Run**
- CGO_ENABLED=0 go vet ./... -> exit 0, clean
- gofmt -l . -> no output, clean
- CGO_ENABLED=0 go test ./... -> all packages ok (audio, brain, codecheck, device, feedback, hid, session, speech, timer, vad, wake)
- CGO_ENABLED=0 go test ./internal/brain/... -v -run 'TestAnAmbiguousDeviceNameAsksInsteadOfGuessing|TestDeviceIDsWithPunctuationStillMatch|TestShippedDefaultAnswersTheClockWithoutAModel' -> all 3 PASS
- go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... -> exit 0, no findings
- go build/test -overlay mutant: weakened rules.go:48 guard from len(named)>1 to len(named)>99 via go-build-overlay (no repo files touched, git status --porcelain empty before/during/after) -> mutant compiles cleanly, then TestAnAmbiguousDeviceNameAsksInsteadOfGuessing FAILS with 'acted on "lamp-2" despite two devices matching' -> mutant killed
- go-build-overlay probe test, inv=[lamp, lamp-2], input 'turn on lamp 2', at head 987ebe0 -> Speak='I know more than one device by that name: lamp, lamp-2. Which one?', Actions=[]
- same probe test, same inventory/input, overlaying rules.go from prior head 07af195 -> Speak='Okay, on lamp.', Actions=[{Device:lamp,Op:on}] (silent wrong-device action, confirms f8's guard is a net improvement not a regression)
- git diff 07af195..987ebe0 -- internal/wake/match.go internal/timer/parse.go internal/vad/energy.go internal/config/config.go internal/audio/earcon.go internal/feedback/feedback.go | wc -l -> 0
- CGO_ENABLED=1 go test -race ./internal/brain/... -> build failed, 'cc1: fatal error: /usr/include/stdc-predef.h: Permission denied'; confirmed environment-wide via raw `gcc -E -x c /usr/include/stdc-predef.h -o /dev/null` (same error) - a sandbox/toolchain limitation unrelated to this code, not a code failure; race detector could not be independently executed here

jerryfane:
## Merged as `a209aab`; `#31-f9` and `#31-f10` deferred to issue #33

Approved at exact head `987ebe072b348c665a004b7d20344230a0f275b0`, which was still the live head, CLEAN, with CI green — all verified before merging rather than from the notification.

Five rounds, each finding something real:

| Round | Finding |
|---|---|
| 1 | false local answers: "what time does the shop close" answered with the clock |
| 2 | thinking-sound race: fired on 1/500 fast-path answers |
| 3 | no punctuation handling, so the device's own `What time is it?` missed the fast path entirely |
| 4 | device-matching regression **I introduced in round 3** — punctuated IDs stopped matching |
| 5 | ID collision: `lamp-2` and `lamp_2` reduced alike, first match won silently |

Round 3 is the one worth keeping: the reviewer used the device's own journal line, `heard="Hey boys. What time is it?"`, and my tests had all passed on tidy strings I invented. The fast path would have shipped doing nothing.

`#31-f9` and `#31-f10` are in issue #33 with acceptance criteria — prefer the longest ID match rather than asking needlessly, and speak IDs without punctuation so the clarification can actually be answered out loud.

Deploying to the device now. I will report the measured `stage=plan` duration for a clock query rather than claiming it got faster.

PR #34, opened 2026-09-18
feat(feedback): settable thinking volume, and record spoken requests instead of losing them

jerryfane:
Two things the owner asked for. The second exists because the first was lost.

## `feedback.thinking_volume`

Its own `0`–`1` setting for the thinking sound, previously fixed at a third of `feedback.volume` inside the renderer.

- **unset** → derives from `feedback.volume` at the same third, so an existing config hears exactly what it heard before
- **set** → used as given, including louder than the acknowledgement; it is the owner's room
- **`0`** → silent

One resolver answers the level for the renderer, the validator and `voice doctor`, because two places deriving one number is how a report drifts from behaviour — this repo has been caught doing that twice (`#23-f9`, `#31-f6`).

### Why a field and not a comment

The same knob was asked for twice and I answered neither properly:

1. the reviewer raised it as `#31-f4`, and I replied with a comment
2. the owner asked the **device**, which answered *"I can't adjust the thinking sound volume from the controls available to me."*

That refusal was correct — the agent runs as a restricted account with no checkout — and it went nowhere.

### Zero had to mean silent

Caught in review before shipping: `ThinkingVolume` was a plain `float64`, so an explicit `0` was indistinguishable from an omitted field. Asking for silence, from the setting added to control the volume, would have been read as "derive from `Volume`" and ignored. It is a `*float64` now; the test asserts silence survives to the **rendered audio**, not just out of the resolver, and that a config asking for silence still validates.

## Spoken requests stop vanishing

`internal/requests` records what the voice agent cannot do itself, appended to a file in the workspace it already owns. **No new privileges**: no repo access, no credentials, no `pi` account.

`packaging/AGENTS.md` now tells it to record the request **in the user's own words before answering**, and say it has done so. `voice doctor` reports what is waiting:

```
requests     INFO 1 spoken request(s) need someone with code access:
                  18 Sep 16:05  make the thinking sound volume configurable
```

Verified end to end by running the real `doctor` against a file at the **default** path with no override — the same path the service resolves under `HOME=/home/voice-agent` with `WorkingDirectory=/home/voice-agent/voice-workspace`, which is where the agent writes.

### Design choices worth arguing with

- **verbatim, not paraphrased** — a paraphrase is my reading of the words, and the point is that the request survives without one
- **appended, not rewritten** — a crash loses at most the line being written
- **a missing file is not an error** — no requests and no file are the same state, and `doctor` must not report a problem on a healthy install
- **a malformed line is kept** — a line nobody can parse is still something a person asked for; dropping it loses exactly what the file exists to keep

## Mutants, all failing

| Mutant | Failure |
|---|---|
| resolver ignores the ratio | `unset thinking volume = 0.6, want 0.21` |
| zero treated as unset | `explicit thinking_volume 0 resolved to 0.315; it must mute` |
| loader drops untimestamped lines | `loaded 2 requests, want 3` |

## Verification

Statuses captured directly rather than through a pipeline: `test=0 vet=0 errcheck=0 gofmt=[]`, plus `-race` green on the changed packages.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d66ac45d55ab84-1`

**Summary:** Reviewed jerryfane/voice PR #34 at exact head 5475944e60a200d54ea28b5bc55e67582aed54b7 (base a209aab, confirmed via GitHub API; no prior verdicts exist for this PR). The thinking_volume resolver, pointer semantics, and zero-mutes-silence behavior are all correct and well-tested: ResolvedThinkingVolume() is the single call site used by both the validator and the renderer, an explicit 0 renders zero PCM samples end-to-end, and *float64/omitempty genuinely round-trips (explicit 0 survives Save/Load, an older-binary config missing the key decodes as unset) — all confirmed with executed scratch tests, and all three named mutants (resolver ignoring the ratio, zero treated as unset, loader dropping untimestamped lines) were manually reintroduced in a scratch copy and caught by the existing suite. Build/vet/test/gofmt/errcheck all pass with directly-captured exit codes, and CI's own 'test' check-run on this exact head is confirmed SUCCESS via the GitHub API. -race could not be run here: this sandbox's AppArmor confinement denies even root read access to /usr/include, blocking cgo entirely (verified independent of the tool's own sandbox flag); CI itself doesn't run -race either. However, attacking the requester's stated top priority — whether the reader (`voice doctor`) and writer (the OMP agent via packaging/voice-agent-run + AGENTS.md) genuinely agree on REQUESTS.tsv's location under the shipped unit — found that they do not for the realistic, documented invocation: built the real binary and ran doctor end-to-end under several HOME/VOICE_AGENT_WORKSPACE combinations. voice.service sets only HOME=/home/voice-agent (no VOICE_AGENT_WORKSPACE), and neither install-local.sh nor README ever arrange for or document running `voice doctor` under that same HOME — the documented `voice doctor` invocation runs as the operator's own login. When the wrapper's VOICE_AGENT_WORKSPACE is overridden (explicitly named as a case to check), doctor didn't just show nothing — it displayed a stale
[truncated]
**Severity:** `P1`

**Findings**
- **voice doctor and the OMP agent's writer can silently disagree on REQUESTS.tsv's location under the shipped deployment** (P1)
  - evidence: cmd/voice/main.go requestsPath() (lines ~71-87) falls back to os.UserHomeDir()+"/voice-workspace" when VOICE_REQUESTS/VOICE_AGENT_WORKSPACE are unset. packaging/voice.service sets only Environment=HOME=/home/voice-agent (no VOICE_AGENT_WORKSPACE); packaging/voice-agent-run defaults the agent's workspace to ${VOICE_AGENT_WORKSPACE:-$HOME/voice-workspace} under HOME=/home/voice-agent (via `sudo -u voice-agent`). Neither packaging/voice.service, scripts/install-local.sh, nor README.md's documented `voice doctor` usage ever runs doctor with HOME=/home/voice-agent — the sudoers grant install-local.sh writes only permits `voice-agent-run *`, not an arbitrary `voice doctor` as voice-agent. Reproduced by building the binary (CGO_ENABLED=0 go build ./cmd/voice) and running doctor with HOME set to a simulated admin account (mirroring the documented invocation): the requests block was silently absent, no error. Reproduced the requester's named edge case (wrapper's VOICE_AGENT_WORKSPACE overridden) and found the failure mode is worse than absence: doctor, run under the shipped unit's HOME with no override, displayed a STALE prior request as if it were the complete current report while a newer request sat unseen in the overridden workspace — verified by inspecting both files' actual contents after the run.
- **Malformed journal lines are indistinguishable from well-formed ones in voice doctor, and the real writer bypasses Append()'s sanitization** (P2)
  - evidence: cmd/voice/main.go doctor(): every entry prints as `undated  <text>` or `<date> <text>` with no distinct malformed marker, whether the line lacked a tab, had an unparseable timestamp, or is a genuine untimed entry. internal/requests/requests.go's line()/strings.Fields sanitization (collapses embedded tabs/newlines) is exercised only by requests_test.go — the real production writer is an OMP agent following packaging/AGENTS.md's prose instruction, not this Go code. Reproduced: writing a raw file with an embedded newline inside one spoken request (`ts\tplay music\nthen turn the volume down\n`) causes Load() to return two separate entries — one truncated to 'play music', one zero-time 'undated' entry showing 'then turn the volume down' — both rendered by doctor as if they were two independent, legitimate requests.
- **AGENTS.md's recording instruction doesn't say verbatim** (P3)
  - evidence: packaging/AGENTS.md: '...then the request in the user's own words.' Does not use 'verbatim', 'exact quote', or 'do not paraphrase/summarize', despite requests.go's own doc comment stating verbatim survival without a paraphrase is the entire point of the package. Not independently testable here (no OMP agent execution available); flagged per explicit reviewer instruction to assess only the wording.
- **Stale doc reference to a nonexistent `voice requests` command** (P3)
  - evidence: internal/requests/requests.go package doc: "the file lives in the workspace it already owns, and `voice requests` reads it back." No such subcommand exists anywhere in cmd/voice/main.go in this diff or base; the actual reader is `voice doctor`.

**Tests Run**
- cd workspace && CGO_ENABLED=0 go build ./... — exit 0
- cd workspace && CGO_ENABLED=0 go vet ./... — exit 0
- cd workspace && CGO_ENABLED=0 go test ./... — exit 0, all 15 packages ok (including internal/config, internal/requests, internal/audio, internal/session)
- cd workspace && gofmt -l <changed files> and gofmt -l . — exit 0, no output on either (clean)
- cd workspace && CGO_ENABLED=0 go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... — exit 0, no findings (matches CI's exact invocation)
- go test ./internal/config/... ./internal/requests/... ./internal/audio/... ./internal/session/... -race — FAILED TO RUN: 'go: -race requires cgo; enable cgo by setting CGO_ENABLED=1', and CGO_ENABLED=1 builds fail in this sandbox with 'fatal error: /usr/include/stdc-predef.h: Permission denied' even as root and even with the tool's own sandbox disabled (confirmed via `id`, `stat`, apparmor module loaded) — an environment-level restriction, not something addressable here; CI's own ci.yml also runs `go test ./...` without -race
- curl https://api.github.com/repos/jerryfane/voice/commits/5475944.../check-runs — confirms the 'test' check-run status=completed conclusion=success on the exact PR head
- curl https://api.github.com/repos/jerryfane/voice/pulls/34 — confirms base=a209aab5071f67eca409e2d2dcdb921193c36bb5, head=5475944e60a200d54ea28b5bc55e67582aed54b7, state=open, mergeable_state=clean; 0 review comments, 0 issue comments
- Scratch copy mutation testing (3 mutants, each reverted after): (1) ResolvedThinkingVolume() made to ignore the ratio and always return Volume — caught by TestThinkingVolumeIsSettableAndDefaultsToAShareOfVolume and TestShippedDefaultKeepsTheExistingThinkingLevel; (2) resolver changed to treat an explicit 0 as unset (falsy check instead of nil check) — caught by TestAnExplicitZeroMutesTheThinkingSound; (3) requests.Load() changed to drop no-tab lines instead of keeping them — caught by TestAMalformedLineIsKeptRatherThanDropped. All three ran via `CGO_ENABLED=0 go test ./internal/config/...` / `./internal/requests/...` and failed as expected
- Scratch Go test: JSON round-trip of Feedback.ThinkingVolume — explicit 0 survives MarshalIndent (`"thinking_volume": 0` present, not omitted) and Unmarshal (`*ThinkingVolume == 0`, not nil); a config missing the key entirely (simulating an older binary) decodes as nil (unset) — PASS
- Scratch Go tests on internal/requests, run via `CGO_ENABLED=0 go test ./internal/requests/... -run TestScratch -v`: embedded-tab raw write preserved the tab in Text (no data loss, contrary to initial hypothesis); embedded-newline raw write split into two doctor-visible entries (one truncated, one fabricated-looking 'undated' fragment) — confirms finding P2; 100 concurrent in-process Append() calls from two goroutines round-tripped with no loss; a 200000-byte line round-tripped intact; Append() against an uncreatable workspace (parent path is a file) returned a clean error with no partial write
- End-to-end: built `voice` binary (CGO_ENABLED=0 go build -o $TMPDIR/voice ./cmd/voice), generated a config via `voice --config PATH init --force`, and ran `voice --config PATH doctor` under `env -i HOME=<home> PATH=<path>` for: (a) HOME matching voice.service's Environment=HOME=/home/voice-agent with a request written to that exact default workspace — request correctly found and printed (matches the operator's own claimed positive test); (b) HOME set to a different simulated admin account with no override, same request present — requests block silently absent, no error, no hint anything was missed; (c) writer using VOICE_AGENT_WORKSPACE override to a separate directory while the reader (matching the shipped unit) has no override — doctor displayed a stale prior request as the complete report while the newer request in the override workspace went unseen

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d66bedc200cfa4-1`

**Summary:** Reviewed jerryfane/voice PR #34 at exact head ad4d4fa7b821b8dbaec231c14da6e65ce3d6bc33, diffed against prior head 5475944e60a200d54ea28b5bc55e67582aed54b7. F2, F3, F4 are resolved and verified by execution/reading. F1's default-install case is genuinely fixed (requestsPath now resolves the voice-agent account's real home via a testable os/user seam, verified with a passing, non-skipping test on an account-less host) but I reproduced a narrower residual divergence: a workspace relocated solely via voice.service's Environment=VOICE_AGENT_WORKSPACE is honoured by the writer (inside that unit) but invisible to a manually-run `voice doctor`, which reports nothing at all with no error while a real request sits unseen - downgraded to P2 since it now needs a deliberate, undocumented non-default operator action to trigger, but it is the same silent-wrong-answer class the requester flagged as the one that matters, so not closed. Build/vet/gofmt clean on this exact head with directly-captured exit codes; full test suite for the two changed packages passes (8/8); CI's own 'test' check-run confirmed SUCCESS via the GitHub API. errcheck was not available in this environment so that specific claim is unverified rather than confirmed. Same-model-family caveat: this review, the prior PR #34 verdict, and (per commit history) the implementation all appear to be Claude-family agent work.
**Severity:** `P2`

**Findings**
- **Default-install divergence is fixed; a narrower silent divergence remains when the workspace is relocated only via voice.service's Environment=** (P2)
  - continues_uid: jerryfane/voice#34-f1
  - detail: Verified the fix: requestsPath() now resolves the voice-agent account's real home via a user.Lookup seam (agentHome), independent of the caller's own HOME, with VOICE_REQUESTS/VOICE_AGENT_WORKSPACE still winning outright and a caller-home fallback only when no such account exists. Ran `CGO_ENABLED=0 go test ./cmd/voice/... -run TestRequestsPath -v` - all 3 tests pass, including TestRequestsPathResolvesTheAgentWorkspaceNotTheCallers, which does NOT skip on this account-less host (confirmed by reading the seam and by the passing run), closing the requester's self-flagged 'first test skipped and asserted nothing' gap. I could not literally rebuild the described mutant (requestsPath reverted to os.UserHomeDir) because this job's sandbox is read-only everywhere except $TMPDIR, so that specific claim is verified by code-reading plus the passing real test, not by an executed mutation. Separately, I reproduced a residual gap: built the real voice binary, wrote a REQUESTS.tsv to a 'relocated' workspace path, then ran `voice doctor` with VOICE_AGENT_WORKSPACE unset (simulating an operator's own shell, distinct from voice.service's process-scoped Environment=) - doctor printed no requests line at all, no error, no FAIL, while the real request sat unreported. This happens because voice-agent-run's internal `sudo -n -H` re-exec is bypassed entirely in the standard systemd path (the unit already runs as User=voice-agent, so the id-check is false and env flows through unstripped to the writer), but voice.service's Environment= directives are scoped to that unit's own process tree and never reach a separately-invoked `voice doctor`. Recommendation: don't have the reader ignore VOICE_AGENT_WORKSPACE (that would break the already-tested single-shell override workflow); instead have doctor additionally consult the packaged unit's configured environment (e.g. `systemctl show voice.service --property=Environment`) when the shipped deployment is present, and print a loud warning 
[truncated]
  - file: cmd/voice/main.go:81-119, packaging/voice.service:8-15, packaging/voice-agent-run:3-7
  - state: answered
- **Resolved: malformed lines are now marked 'unparsed' and shown rather than hidden** (P2)
  - continues_uid: jerryfane/voice#34-f2
  - detail: Verified by execution, not just reading. `CGO_ENABLED=0 go test ./internal/requests/... -v` passes, including TestAMalformedLineIsKeptRatherThanDropped which confirms both a no-tab line and an unparseable-timestamp line are kept as text rather than dropped. Additionally built the real voice binary and ran `voice doctor` against a synthetic REQUESTS.tsv with a dated line, an unparseable-timestamp line, and a no-tab line: output showed the dated line with a real timestamp ('18 Sep 15:09'), and both malformed lines marked 'unparsed' and still printed with their full text - confirming the marker is textually distinct from the date format used ('2 Jan 15:04') and cannot be confused with a timestamp, and that a genuinely untimed line is shown rather than silently dropped. Append()'s sanitization is still only exercised by tests, not by the real OMP-agent writer, but doctor's display-side fix correctly assumes any zero-time entry is malformed relative to AGENTS.md's contract (every legitimate line always carries a real RFC3339 stamp), so this is not a gap in practice.
  - file: cmd/voice/main.go:315, internal/requests/requests.go:83-99, internal/requests/requests_test.go:75-96
  - state: answered
- **Resolved: AGENTS.md now instructs verbatim recording on one line, with rationale** (P3)
  - continues_uid: jerryfane/voice#34-f3
  - detail: packaging/AGENTS.md now reads: 'the request VERBATIM in the user's own words, on ONE line. Do not paraphrase, summarise or translate it, and do not include a newline inside it: a paraphrase is your reading of what was said rather than what was said, and an embedded newline splits one request into two.' Judged as an instruction to a model: unambiguous, uses explicit negative instructions (do not paraphrase/summarise/translate, no embedded newline), and states the why for both failure modes previously reproduced (paraphrase, and an embedded newline splitting one request into two doctor-displayed entries). Not independently testable here since no OMP agent execution is available in this environment; this is a static wording judgment only, as before.
  - file: packaging/AGENTS.md:11
  - state: answered
- **Resolved: stale 'voice requests' doc reference removed** (P3)
  - continues_uid: jerryfane/voice#34-f4
  - detail: internal/requests/requests.go's package doc now reads 'the file lives in the workspace it already owns, and `voice doctor` reads it back,' matching the actual reader subcommand. Confirmed via diff against prior head.
  - file: internal/requests/requests.go:13
  - state: answered

**Tests Run**
- CGO_ENABLED=0 go test ./cmd/voice/... ./internal/requests/... -count=1 -v -> ok, 8/8 tests PASS (TestRequestsPathResolvesTheAgentWorkspaceNotTheCallers, TestRequestsPathFallsBackWhenTheAccountIsAbsent, TestRequestsPathHonoursExplicitOverrides, TestARecordedRequestSurvivesVerbatim, TestRequestsAccumulateInOrder, TestNoFileIsNotAnError, TestAMalformedLineIsKeptRatherThanDropped, TestAnEmptyRequestIsRefused)
- gofmt -l cmd/voice/main.go cmd/voice/requests_path_test.go internal/requests/requests.go -> no output (clean)
- Built real ./cmd/voice binary (CGO_ENABLED=0) into a writable scratch dir; ran `voice init` then `voice doctor` against a synthetic REQUESTS.tsv with dated + unparseable-timestamp + no-tab lines via VOICE_AGENT_WORKSPACE override -> requests block correctly showed 1 dated entry and 2 'unparsed' entries, none dropped
- Reproduced residual F1 gap: wrote REQUESTS.tsv to a 'relocated' workspace, ran `voice doctor` on an account-less host with VOICE_AGENT_WORKSPACE unset (simulating an operator shell distinct from voice.service's own process env) -> doctor printed no requests block at all, no error, while the file held a real unreported request
- curl https://api.github.com/repos/jerryfane/voice/commits/ad4d4fa7b821b8dbaec231c14da6e65ce3d6bc33/check-runs -> 'test' check-run status=completed, conclusion=success (unauthenticated GitHub API, corroborates requester's CI claim)
- errcheck: binary not present in this environment; not run, not claimed

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d66ce2c8d99e70-1`

**Summary:** At head 993b9e9, requestsPath() is gone and requestsPaths() now returns every candidate (VOICE_REQUESTS, relocated VOICE_AGENT_WORKSPACE, the voice-agent account's real home, caller's-home backstop), deduplicated by string, most specific first; doctor reads all of them, sums the total, and names the containing file whenever more than one location holds requests. This closes the residual F1 gap I reproduced last round (a workspace relocated only via voice.service's Environment= was invisible to a manually-run doctor) — confirmed by building the real binary and running doctor against two distinct real paths, each holding one request: both were found, correctly labeled with their path, and the total summed to 2; a single-location case correctly suppressed the path label. All 5 new tests pass and none pass for the wrong reason on inspection: the 'always include agent workspace' and 'both sides of relocation' tests assert inclusion via hasPath (not exhaustiveness, matching their stated scope); the override-first test only pins index 0; the dedup test exercises all three add() calls collapsing to one identical string; the never-empty test correctly exercises HOME='' driving os.UserHomeDir() to error before falling to the relative-path last resort. Full suite (go build, go vet, gofmt, go test ./..., and the repo's own errcheck-equivalent codecheck test) is clean. However, exercising the requester's own first stated risk found a real, reachable new defect: because dedup compares strings only, two candidates that resolve to the same underlying file via a symlink or bind mount (e.g. VOICE_AGENT_WORKSPACE set to a bind-mounted alias of the agent's real home, a realistic pattern on Pi-class hardware with external storage) are NOT deduplicated — reproduced by execution: the same request was counted twice (total=2 for one real request) and printed twice under two different path labels. This is new behaviour: the prior single-path resolver could never double-report since only one 
[truncated]
**Severity:** `P2`

**Findings**
- **Relocation-visibility gap is fixed; string-only dedup lets a symlinked/bind-mounted duplicate candidate double-count and double-print the same request** (P2)
  - continues_uid: jerryfane/voice#34-f1
  - detail: Verified by execution that the core residual gap from ad4d4fa is closed: built the real voice binary, wrote distinct REQUESTS.tsv files to two real candidate locations (a relocated VOICE_AGENT_WORKSPACE and the caller's-home backstop, standing in for the account-based agent home since this sandbox has no voice-agent account) and ran `voice doctor` with VOICE_AGENT_WORKSPACE set but no other propagation trick: both files were found, each correctly labeled 'in <path>', and the total correctly summed to 2 (`CGO_ENABLED=0 go build -o $TMPDIR/voice ./cmd/voice` then `voice doctor` against $TMPDIR/e2e/{home/voice-workspace,relocated}/REQUESTS.tsv`). Removing the second file and re-running confirmed the single-location case correctly suppresses the path label (no noise). `CGO_ENABLED=0 go test ./cmd/voice/... -run TestRequestsPaths -v` passes all 5 new tests; `CGO_ENABLED=0 go build ./...`, `go vet ./...`, `gofmt -l .` are all clean; `go test ./...` is green across every package including `internal/codecheck`'s own errcheck-equivalent (TestNoUnjustifiedDiscardedResults, 48 files scanned, none unjustified). However, the requester's own self-flagged risk is real and reachable: I created a directory `real/` with one request, symlinked both `home/voice-workspace -> real` and `relocated_link -> real`, set HOME=.../home and VOICE_AGENT_WORKSPACE=.../relocated_link, and ran doctor - it reported 'total=2' and printed the identical request text and timestamp twice, once under each symlinked path label, because requestsPaths()'s dedup (`cmd/voice/main.go:97-107`, the `add` closure) compares raw strings, not resolved identity. A workspace relocated via a symlink or bind mount to the same underlying directory as the agent's real home - a normal pattern for external storage on embedded hardware - would silently double-count and double-print every pending request. Recommend resolving each candidate (filepath.EvalSymlinks, or os.SameFile on the Stat result) before the equality ch
[truncated]
  - file: cmd/voice/main.go:97-119
  - state: open
- **An unreadable candidate at the lowest-priority caller's-home backstop fails the whole doctor run even when the real agent journal is healthy and empty** (P3)
  - continues_uid: jerryfane/voice#34-f1
  - detail: Reproduced by execution, answering the requester's direct question: with no real journal present (agentHome/VOICE_AGENT_WORKSPACE candidates correctly returned nil,nil - a missing file is not an error per requests.Load) but a directory literally named REQUESTS.tsv left over in the caller's own home (the lowest-priority backstop candidate), `voice doctor` printed `requests FAIL <path>: read ...: is a directory` and no INFO line at all, even though the device's actual agent-managed journal has nothing wrong with it. This is a judgment call the requester explicitly invited pushback on, not a hard defect: treating any unreadable candidate as a fault is defensible given the whole feature exists to avoid silently trusting one location, but the backstop candidate is speculative developer-machine infrastructure, not the real device's journal, so failing an otherwise-healthy install over garbage sitting there specifically seems more aggressive than warranted. Consider not hard-failing on the caller's-home backstop specifically, or documenting the tradeoff.
  - file: cmd/voice/main.go:317-327
  - state: open
- **Still resolved: malformed-line marking is unchanged by this diff** (P3)
  - continues_uid: jerryfane/voice#34-f2
  - detail: Confirmed via `git diff ad4d4fa..993b9e9` that internal/requests/requests.go was not touched by this head; the 'unparsed' display logic in doctor's print loop is preserved verbatim, just moved one level deeper to iterate per matched path. `go test ./internal/requests/...` remains green (part of the full `go test ./...` run).
  - file: cmd/voice/main.go:337-343, internal/requests/requests.go:83-99
  - state: answered
- **Still resolved: AGENTS.md verbatim-recording instruction is unchanged by this diff** (P3)
  - continues_uid: jerryfane/voice#34-f3
  - detail: packaging/AGENTS.md is not in the changed-files list for this head and the diff confirms it; no regression possible. Not independently re-testable here (no OMP agent execution available), same as the prior round.
  - file: packaging/AGENTS.md:11
  - state: answered
- **Still resolved: package doc reference is unchanged by this diff** (P3)
  - continues_uid: jerryfane/voice#34-f4
  - detail: internal/requests/requests.go was not touched by this head; the corrected package doc referencing `voice doctor` as the reader remains in place.
  - file: internal/requests/requests.go:13
  - state: answered

**Tests Run**
- CGO_ENABLED=0 go build ./... -> ok
- CGO_ENABLED=0 go vet ./... -> ok, no findings
- gofmt -l . -> clean, no files listed
- CGO_ENABLED=0 go test ./cmd/voice/... -run TestRequestsPaths -v -> PASS x5 (TestRequestsPathsAlwaysIncludeTheAgentWorkspace, TestRequestsPathsCoverBothSidesOfARelocation, TestRequestsPathsPutAnExplicitOverrideFirst, TestRequestsPathsAreDeduplicated, TestRequestsPathsNeverEmpty)
- CGO_ENABLED=0 go test ./... -> ok across all 16 packages
- CGO_ENABLED=0 go test ./internal/codecheck/... -run TestNoUnjustifiedDiscardedResults -v -> PASS, 48 Go files scanned, no unjustified discards
- CGO_ENABLED=0 go build -o $TMPDIR/voice ./cmd/voice; then real-binary `voice doctor` runs: (1) two distinct real files -> both found, labeled, total=2; (2) single real file -> path label correctly suppressed; (3) two candidates symlinked to one real file -> same request double-counted (total=2 for 1 real request) and printed twice, reproducing the requester's self-flagged dedup risk; (4) one EISDIR candidate + one healthy candidate -> FAIL line for the bad candidate, INFO line still correct for the healthy one; (5) EISDIR at the caller's-home backstop with the real journal absent (healthy/empty) -> whole run reports FAIL, no INFO line, on an otherwise-healthy device

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d66ddc99e68fb9-1`

**Decision:** `approved`
**Summary:** At head 619d945, both open findings from the 993b9e9 review are fixed and independently verified by execution against the real binary, not just by reading the diff. (1) Dedup now compares file identity (os.Stat+os.SameFile) with a string-equality fallback only for not-yet-existing candidates; a real symlinked-directory relocation (agent home's voice-workspace symlinked to the same dir as a relocated VOICE_AGENT_WORKSPACE) now reports total=1, not 2. (2) Candidates now carry a required flag; an unreadable directory-named REQUESTS.tsv at the caller's-home backstop no longer fails doctor (no 'requests FAIL' line) while a genuine request elsewhere is still reported correctly, and the backstop is surfaced as 'requests INFO ignoring an unreadable backstop: ...' with the underlying path still visible through the wrapped os.PathError text. Build, vet, gofmt, and the full test suite (including internal/codecheck's discarded-result check) are clean. Requester's three specific technical questions: os.Stat (not Lstat) is the correct call for a symlinked workspace directory - confirmed by a standalone probe that os.Stat correctly follows both an intermediate symlinked directory and a symlinked final file to the same identity, while Lstat on a symlinked final file would break the match; a not-yet-written candidate cannot be stat'd so it falls back to string dedup, but this is harmless because a missing file contributes no data to double-count. The bind-mount st_dev/st_ino claim could not be executed here - this sandbox denies mount(2) even as uid 0 (no user-namespace/CAP_SYS_ADMIN effect available, 'permission denied' on both a plain bind mount and unshare -m) - so that specific sub-claim rests on documented Linux semantics (a bind mount is a second vfsmount over the same inode, not a new filesystem instance, so stat() returns identical st_dev/st_ino) rather than a syscall I performed; flagging this per the instruction to say so rather than claim executed for what I could not run
[truncated]

**Findings**
- **Relocation-visibility gap is fixed; dedup now compares file identity, closing the symlinked/bind-mounted double-count** (P2)
  - continues_uid: jerryfane/voice#34-f1
  - detail: Verified by execution. Built the real binary (CGO_ENABLED=0 go build -o $TMPDIR/voice ./cmd/voice) and reproduced the exact prior repro: real/REQUESTS.tsv with one request, home/voice-workspace and relocated_link both symlinked to real/, HOME set to the home dir and VOICE_AGENT_WORKSPACE set to relocated_link. `voice doctor` now reports `requests INFO 1 spoken request(s)` (previously 2), confirming os.Stat+os.SameFile-based identity dedup in cmd/voice/main.go's `add` closure collapses the duplicate. Also ran `go test ./cmd/voice/... -run TestRequestsPaths -v` (6 tests, all PASS, including the new TestRequestsPathsDeduplicateBySymlinkedIdentityNotJustString using a real os.Symlink) and the full `go test ./...` (all packages ok). Independently probed the three technical sub-questions the requester asked about: (a) a standalone Go program confirms os.Stat correctly resolves identity through both a symlinked intermediate directory and a symlinked final file, matching SameFile=true in both cases, while Lstat on a symlinked final file gives SameFile=false - so os.Stat (not Lstat) is the right call, as shipped. (b) a not-yet-existing candidate behind a symlinked dir causes os.Stat to error, which falls back to string-only comparison in the `add` loop - this is safe because a nonexistent file yields no data from requests.Load to double-count, so the two un-deduped string entries never produce duplicate output. (c) the bind-mount st_dev/st_ino claim could not be executed: this sandbox refuses mount(2) even as uid 0 ('permission denied' on a plain bind mount and on `unshare -m`, with no working user namespace), so I could not literally verify a bind mount produces matching SameFile results here; it rests on documented Linux semantics (a bind mount is a second vfsmount over the same inode/superblock, not a distinct filesystem instance, so stat() returns identical st_dev/st_ino) rather than an executed syscall - flagging this gap in evidence mode rather than claiming it
[truncated]
  - file: cmd/voice/main.go:106-136
  - state: answered
- **Backstop severity judgment (INFO, not FAIL) is a reasonable call and the wording is clear enough in practice** (P3)
  - continues_uid: jerryfane/voice#34-f1
  - detail: Verified by execution. Built agentDir/voice-workspace/REQUESTS.tsv holding a genuine request and a directory literally named REQUESTS.tsv under a separate caller's-home/voice-workspace (the backstop). Ran `voice doctor` with VOICE_AGENT_WORKSPACE pointed at the real journal and HOME pointed at the poisoned backstop: output shows `requests INFO ignoring an unreadable backstop: read requests: read <path>: is a directory` followed by `requests INFO 1 spoken request(s) ... play music`, and `grep -i FAIL` on the output has no `requests` line - confirms the healthy real journal is unaffected and the backstop no longer hard-fails doctor. On the requester's specific question of whether the phrasing could be misread as the device ignoring a real request: the message names both the reason ('unreadable') and the role ('backstop') explicitly, and the underlying os.PathError text (embedded via %v) still names the actual failing path, so a reader has enough information to distinguish 'a secondary location I can't read' from 'a request I chose to drop'. I don't see a stronger phrasing needed; no pushback here. Separately confirmed the required-candidate failure path still surfaces the path for debugging even after requests.Load's message was simplified: a directory at VOICE_REQUESTS (required) produced `requests FAIL read requests: read <full path>: is a directory`, so nothing was lost by dropping the old triplicated `path+err.Error()` format, and no test or caller in the repo asserts on the old message shape.
  - file: cmd/voice/main.go:344-358
  - state: answered
- **Still resolved: malformed-line marking is unchanged by this head** (P3)
  - continues_uid: jerryfane/voice#34-f2
  - detail: `git diff 993b9e9..619d945 -- internal/requests/requests.go` shows only the Load() error-wrap line changed (`fmt.Errorf("read %s: %w", path, err)` -> `fmt.Errorf("read requests: %w", err)`); the per-line 'unparsed' marking logic in both the loader and doctor's print loop is byte-identical to the prior round, just indented one level deeper to iterate per matched candidate. `go test ./internal/requests/... ` and the doctor print path both pass as part of the full `go test ./...` run.
  - file: cmd/voice/main.go:376-382, internal/requests/requests.go:83-99
  - state: answered
- **Still resolved: AGENTS.md verbatim-recording instruction is unchanged by this head** (P3)
  - continues_uid: jerryfane/voice#34-f3
  - detail: `git diff 993b9e9..619d945 --stat` shows only cmd/voice/main.go, cmd/voice/requests_path_test.go, and internal/requests/requests.go changed; packaging/AGENTS.md is untouched, so no regression is possible. Not independently re-testable here (no OMP agent execution available), same as the prior round.
  - file: packaging/AGENTS.md:11
  - state: answered
- **Still resolved: package doc reference is unchanged by this head** (P3)
  - continues_uid: jerryfane/voice#34-f4
  - detail: Read internal/requests/requests.go:1-13 at this head: the package doc still correctly states '`voice doctor` reads it back'. The diff for this file touches only the Load() error string (line 82), confirmed via `git diff 993b9e9..619d945 -- internal/requests/requests.go`.
  - file: internal/requests/requests.go:13
  - state: answered

**Tests Run**
- CGO_ENABLED=0 go build ./... (BUILD_OK)
- CGO_ENABLED=0 go vet ./... (VET_OK)
- gofmt -l . (clean, no output)
- CGO_ENABLED=0 go test ./... (all packages ok, including internal/requests, internal/codecheck)
- CGO_ENABLED=0 go test ./cmd/voice/... -run TestRequestsPaths -v (6/6 PASS)
- CGO_ENABLED=0 go test ./cmd/voice/... -run TestOnlyTheIntendedLocationsAreRequired -v (PASS)
- CGO_ENABLED=0 go test ./internal/codecheck/... -run TestNoUnjustifiedDiscardedResults -v (PASS, 48 files scanned)
- CGO_ENABLED=0 go build -o $TMPDIR/voice-review/voice ./cmd/voice; real-binary run: `voice doctor` against a symlinked-directory relocation duplicate (real/, home/voice-workspace -> real, relocated_link -> real) reporting total=1 not 2
- real-binary run: `voice doctor` with an unreadable directory-named REQUESTS.tsv at the caller's-home backstop plus a genuine request in the real agent workspace - no `requests FAIL` line, backstop reported as INFO, real request still found
- real-binary run: `voice doctor` with a directory at the required VOICE_REQUESTS candidate - confirms `requests FAIL` still names the full path via the wrapped os.PathError
- standalone Go probe (outside repo, in $TMPDIR) confirming os.Stat (not Lstat) correctly resolves identity through a symlinked directory and a symlinked file
- attempted `mount --bind` and `unshare -m` in this sandbox to directly verify bind-mount st_dev/st_ino identity - both denied ('permission denied'/'cannot open /proc/self/uid_map') even as uid 0, so this specific sub-claim is not independently executed here and rests on documented kernel semantics instead

PR #36, opened 2026-09-18
fix(install): stop overwriting wake settings the installer did not author

jerryfane:
The installer clobbered the owner's wake alias on every run, and somebody put it back by hand afterwards. Every time.

## The defect

`scripts/install-local.sh` assigned `wake.phrases` and `wake.fuzz` unconditionally:

```
.wake.phrases = ["hey voice"] | .wake.fuzz = 0 | del(.wake.follow_up) |
```

So each reinstall silently removed an alias the owner had added and reset his fuzz. Breaking a setting and repairing it afterwards is not keeping it — and the repair depended on a person remembering.

The wake block is now written **only when this script created the config**. On an existing install it is left alone, and the script prints what it kept:

```
keeping existing wake phrases: ["hey voice","hey boys"]
```

**Deliberately not special-cased to one phrase.** The defect is an installer overwriting user configuration it does not own; hardcoding the owner's alias would leave the next person's setting to be clobbered.

### Proven with the real filters

Against a config holding two phrases and a non-zero fuzz:

| | `wake` after the filter |
|---|---|
| owner's config before | `phrases:["hey voice","hey boys"] fuzz:0.34` |
| **shipped filter (main today)** | `phrases:["hey voice"] fuzz:0` — alias gone |
| **new filter, existing config** | `phrases:["hey voice","hey boys"] fuzz:0.34` — untouched |
| **new filter, fresh config** | `phrases:["hey voice"] fuzz:0` — documented default |

## `voice wake TRANSCRIPT`

Whether the config *file* contains a phrase was never the question. After an install the question is whether the words somebody says out loud are accepted, and the only way to answer it was to say them and listen.

```
$ voice wake "hey boys"
YES  "hey boys" wakes Voice on "hey voice", with no trailing command
$ voice wake "hey boys turn on the lights"
YES  "hey boys turn on the lights" wakes Voice on "hey voice", command "turn on the lights"
$ voice wake "hey neighbours"
NO   "hey neighbours" would NOT wake Voice; configured phrases: [hey voice hey boys]
exit 1
```

It calls `wake.Match` — the same function `session.go:151` calls — rather than a second copy of the rules, because a checker with its own rules can agree with itself while disagreeing with the device. A non-match exits non-zero so a deploy receipt can assert it rather than read it.

## Known wart, filed not fixed here

`hey boys` reports waking on `"hey voice"`: `Match` returns the first phrase that matches, and with fuzz at 0.34 the fuzzy match on `hey voice` wins before the *exact* match on `hey boys` is considered. Matching is correct, attribution is not. Preferring an exact match over a fuzzy one changes which phrase is credited and therefore how a trailing command is extracted, so it belongs in its own reviewed change rather than riding along in an installer fix.

## Verification

`test=0 vet=0 errcheck=0 gofmt clean`, statuses captured directly. The mutant returning `nil` on a non-match fails, since a check that cannot fail is the recurring defect in this repo.

---

## Two defects I introduced in this fix, both caught by review

Recorded here rather than buried in commits, because both are the same failure mode this repo keeps producing: **the failure and the success looked identical to the script.**

### 1. A pipeline that could not fail, installing a 0-byte config (P1)

I split the config rewrite into two piped `jq` calls. The shebang is `#!/usr/bin/env sh` → dash. Dash has no `pipefail`, and `set -e` in a pipeline inspects only the **last** command's status.

So with an unparseable config — a bad manual edit, an unclean power loss on an embedded device — the first `jq` died, the second emitted nothing, and:

```
$ dash -c 'set -eu; jq "." corrupted.json | jq ".brain.mode = \"agent\"" > out.json; echo REACHED INSTALL STEP'
REACHED INSTALL STEP
   bytes it would have installed: 0
```

The single-`jq` form I had replaced aborted instead. **My fix for a clobbered wake alias could have destroyed the entire config.** It is one `jq` again, with the wake clause composed into the same filter so its exit status *is* the status `set -e` sees. Verified under dash: corrupted input exits 5 and the original file is untouched.

### 2. The guard against that was a second copy of a rule, and had the rule wrong (P2)

I then added a belt-and-braces guard asserting `wake.phrases` is non-empty. But `config.Validate` requires phrases only for the **`stt`** detector — an `external` detector legitimately has none and supplies `wake.command` instead.

So the guard I added to protect a reinstall would have **refused** a reinstall on any device using an external wake detector. Two places encoding one rule is how a check drifts from the behaviour it checks, and I wrote a fresh instance of it into a guard against exactly that class of failure.

The check is now the binary itself — `voice --config FILE config`, which loads and validates — so the installer cannot disagree with the application about what a valid config is.

### Verified under dash with the real binary

| Case | Result |
|---|---|
| config with an added alias | installs, keeps `["hey voice","hey boys"]` and `fuzz 0.34` |
| **`external` detector, no phrases** | installs — the case my guard wrongly refused |
| corrupted JSON | exits 5 at `jq`, original file intact |
| `stt` detector, zero phrases | refused before the install step |


jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d66e401d92e52c-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d66e401d92e52c-1`

**Summary:** Reviewed jerryfane/voice PR #36 at exact head 4338fe9 against base a209aab (1 commit). No prior verdict exists for PR #36 in the frozen feed, so no prior findings to answer. Build/vet/full test suite/gofmt all pass; the wake-preservation golden path (reinstall keeps owner's phrases+fuzz, fresh install still gets defaults) was reproduced by executing the actual script text, and the new `voice wake` command was confirmed to share the exact production wake.Match code path as the listener, including reproducing the disclosed 'hey boys attributes to hey voice' wart. However, converting the single jq call into `sudo jq "$wake_filter" ... | jq '...'` (install-local.sh:70) is a real regression: the script is #!/usr/bin/env sh (dash, which has no pipefail - confirmed 'Illegal option'), so when an existing /etc/voice/config.json is invalid JSON, the first jq fails but the second jq silently succeeds on empty stdin (jq treats 0 input values as success) producing a 0-byte file, which the script then installs over the real config - I reproduced this exactly. The pre-PR single-jq version correctly aborted (exit 5) on the identical corrupted input before touching the file, confirmed by direct execution, so this is new. The script's later PID/hash check will eventually catch the resulting service failure and exit 1 (not a silent false-success), but by then the real config is already destroyed and the error message points at the wrong cause. Everything else attacked (quoting of wake_filter with embedded quotes/brackets in both dash and bash, the sudo-jq-to-non-sudo-jq pipe reading a 0640 file, and the fresh_config signal vs. Go's own defaults-then-decode-merge defense in config.Load for a missing/empty wake block) checked out fine on execution. Scope-creep judgment: the voice wake addition is justified, not blocking, since it's small, tested, and is the only way to demonstrate this PR's own acceptance criterion without speaking aloud. Disclosure: this review was performed by a Claud
[truncated]
**Severity:** `P1`

**Findings**
- **P1**
  - description: The old single `sudo jq '...' /etc/voice/config.json > /tmp/voice-config.json` was replaced with `sudo jq "$wake_filter" /etc/voice/config.json | jq '...' > /tmp/voice-config.json`. The script's shebang is #!/usr/bin/env sh, which resolves to dash on this stack; dash has no pipefail (`set -o pipefail` errors with 'Illegal option -o pipefail', confirmed by direct execution), and `set -eu` in a pipeline only inspects the exit status of the LAST command. Reproduced directly: with an existing /etc/voice/config.json that is invalid JSON (e.g. corrupted by disk fault, unclean power loss, or a bad manual edit - plausible on an embedded device), the first jq fails and prints a parse error to stderr, but the second jq receives immediate EOF on stdin; jq treats zero input values as success rather than error, so it exits 0 having written 0 bytes. The script does not abort and proceeds to `sudo install -o root -g "$agent_user" -m 0640 /tmp/voice-config.json /etc/voice/config.json`, replacing the real config with an empty file. I confirmed the pre-PR single-jq form correctly aborted with exit 5 on the identical corrupted input, before ever writing to /etc/voice/config.json, so this is a regression introduced by this PR and not pre-existing. The script's later binary/PID verification step will eventually detect the resulting voice.service crash and exit 1 (this is not a silently-successful install), but only after the previously-installed config.json (which may have still had recoverable, working content) has been irrecoverably overwritten with 0 bytes, and the error message ('voice.service is not running after restart') does not point at the actual root cause. Minimal fix: split into two separate commands (write the first jq stage to a temp file, let set -e see its real exit status) rather than piping, since dash has no pipefail to lean on.
  - evidence_locator: scripts/install-local.sh:70-79 diff; executed reproduction: dash -c with set -eu, wake_filter='.', identical second-stage filter, against a truncated/invalid-JSON config -> pipeline reaches end-of-script marker (script does not abort) and produces a 0-byte /tmp/voice-config.json; baseline comparison: same corrupted input against the pre-PR single-jq command exits 5 and never reaches the install step.
  - file: scripts/install-local.sh
  - id: pipe-truncation-regression
  - lines: 70-79

**Tests Run**
- CGO_ENABLED=0 go test ./... -> all packages ok, including internal/wake and cmd/voice (new wake_check_test.go)
- gofmt -l cmd/voice/main.go cmd/voice/wake_check_test.go -> no output (clean)
- dash -c 'set -o pipefail; echo ok' and sh -c 'set -o pipefail; ...' -> both fail with 'Illegal option -o pipefail', confirming dash/this script's /bin/sh has no pipefail
- Executed the exact wake_filter string in dash and bash against a config missing .wake entirely, confirming quoting with embedded double-quotes/brackets survives both shells and produces the expected wake object
- Executed the real two-stage pipeline (wake_filter='.' + verbatim second-stage filter) against a valid existing config with two phrases and fuzz 0.34 -> both preserved correctly in merged output
- Executed the real two-stage pipeline against a corrupted (invalid JSON) existing config under set -eu -> pipeline does not abort, produces a 0-byte output file (reproduces the reported defect)
- Executed the pre-PR single-jq command (no pipe) against the identical corrupted config under set -eu -> aborts with exit 5 before writing anything, confirming this is a PR-introduced regression, not pre-existing
- Extracted and ran the actual scripts/install-local.sh lines 51-79 (with sudo shimmed to a passthrough, since real sudo/systemd/useradd are unavailable in this sandbox) against a reinstall scenario with an existing two-phrase/fuzz-0.34 config -> merged intermediate file correctly preserves both phrases and fuzz, matching the PR's claimed evidence
- Built the voice binary (CGO_ENABLED=0 go build ./cmd/voice) and ran `voice wake` against a config with phrases [hey voice, hey boys] fuzz 0.34: 'hey boys' -> exit 0 reporting match on 'hey voice' (reproduces disclosed wart), 'hey voice' -> exit 0 match, 'hey neighbours' -> exit 1 with errWakeNoMatch
- In an isolated copy of the repo (not the reviewed worktree), mutated checkWake to return nil on no-match and reran TestWakeCheckExitStatusDistinguishesMatchFromNoMatch -> test fails as claimed, confirming the mutant-killing property; isolated copy discarded, reviewed worktree confirmed unmodified via git status --porcelain
- In an isolated copy of the repo, added scratch Go tests confirming config.Load populates full Wake defaults (phrases, fuzz, detector) and passes Validate() when the JSON has no .wake key at all, and when it has an empty "wake":{} object -> both pass, showing Go's default-then-decode-merge defends independently of the installer's fresh_config heuristic

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d66f0badb399f2-1`

**Summary:** P1 pipe-truncation regression is fixed and verified: the composed single-jq form correctly aborts (exit 5) under both dash and bash before ever touching /etc/voice/config.json, on a corrupted input, and the golden/fresh paths compose and preserve wake settings correctly in both shells. New P2: the added 'no wake phrases' guard doesn't account for wake.detector=="external" (internal/config/config.go:402-411), which the app's own Validate() explicitly permits with empty phrases, so the installer now wrongly refuses to (re)install a legitimate external-wake-detector configuration.
**Severity:** `P2`

**Findings**
- **Pipe-truncation regression: resolved by collapsing to a single composed jq filter** (P1)
  - continues_uid: jerryfane/voice#36-f1
  - detail: At head 7d09f1f the two piped jq calls were replaced by one jq invocation of a composed $config_filter string, plus a post-hoc content guard. Reproduced directly under dash (env sh resolves to dash on this stack) and bash: feeding a corrupted/unparseable /etc/voice/config.json makes the single jq exit 5 immediately (no pipe to swallow the failure), set -eu aborts the script at that line in both shells, and the destructive `sudo install ... /etc/voice/config.json` step is never reached, so the original file is left untouched. Also verified the composed filter's quoting is correct in both shells: the identity case produces `. |\n  .stt.model = ...` (a valid jq pipe), not a dangling `. |` with nothing after it, and the golden path (existing config) preserves wake.phrases=["hey boys","hey voice"] and fuzz=0.34 while the fresh-install path yields the documented default phrases=["hey voice"], fuzz=0, follow_up deleted. Baseline comparison against the prior (piped) form on identical corrupted input still reaches the install step and produces a 0-byte file, confirming this head's form is the fix.
  - evidence_locator: scripts/install-local.sh:70-91 at 7d09f1f; commands executed: `dash extract.sh corrupted.json 0` -> exit=5, no install reached; `bash extract.sh corrupted.json 0` -> exit=5, identical; `dash/bash extract.sh valid-existing.json 0` -> exit=0, phrases/fuzz preserved; `dash/bash extract.sh fresh-init.json 1` -> exit=0, documented defaults applied; extract.sh is the unmodified lines 58-91 of the reviewed script parameterized only on config path and fresh flag.
  - file: scripts/install-local.sh
  - state: answered
- **New guard rejects a config the application itself considers valid: wake.detector=="external" with empty wake.phrases** (P2)
  - detail: The added guard `jq -e 'type == "object" and (.wake.phrases | length > 0)'` treats any config with an empty wake.phrases array as unfit to install, regardless of wake.detector. But internal/config/config.go:402-411 explicitly allows wake.phrases to be empty when wake.detector=="external" (it only requires wake.command to be non-empty in that mode) - Validate() itself accepts such a config. Reproduced: built a config with wake.detector="external", a non-empty wake.command, and empty wake.phrases (a config that passes the app's own Validate()), ran it through the exact guard logic in this diff, and it was rejected with exit 1 and 'refusing to install a config that is empty, unparseable, or has no wake phrases' - the same message given to a genuinely corrupted config. This means any device legitimately configured for external wake detection cannot be reinstalled or upgraded by this script without the guard being loosened. This directly answers the requester's request to challenge the 'no wake phrases' condition: a phrase-less config is a legitimate, Validate()-accepted configuration under wake.detector=="external", so the guard as written is too narrow. Minimal fix: extend the predicate to `(.wake.detector == "external" and (.wake.command|length)>0) or (.wake.phrases|length>0)`, mirroring config.go's own rule, rather than reimplementing a narrower one-off check in shell.
  - evidence_locator: scripts/install-local.sh new guard block (added at 7d09f1f, immediately after `sudo jq "$config_filter" ... > /tmp/voice-config.json`); internal/config/config.go:402-411 (Validate()'s wake.detector switch, permitting empty Phrases when Detector=="external"); executed reproduction: `dash extract.sh external-detector.json 0` (config: {"wake":{"detector":"external","command":["some-external-wake-detector"],"phrases":[],"fuzz":0},...}) -> exit=1, 'refusing to install...' printed, no output file left.
  - file: scripts/install-local.sh
- **Guard's intermediate /tmp/voice-config.json uses a fixed, predictable path (pre-existing, unchanged by this diff)** (P3)
  - detail: Checked per requester's ask about the guard's unprivileged jq read and the cleanup rm on the failure path. The `> /tmp/voice-config.json` redirection is opened by the calling (unprivileged) shell process itself, not by sudo - only the jq invocation that reads /etc/voice/config.json runs under sudo - so the intermediate file is created owner:owner with the default umask (confirmed 0644 in a live test), not root:agent 0640; the guard's unprivileged `jq -e` read therefore has no permission problem. The fixed non-random /tmp/voice-config.json filename (used identically before and after this diff, including the pre-existing single-jq baseline) is a theoretical symlink-race surface for a local attacker who can predict/pre-create that path, but this is unchanged by the current diff and not a new defect it introduces, so it does not block this review.
  - evidence_locator: scripts/install-local.sh (intermediate file path, unchanged across 4338fe9 and 7d09f1f); executed: `jq '.' valid-existing.json > permtest.json; ls -la permtest.json` -> -rw-r--r-- owner:owner, confirming unprivileged-writable/readable, no permission mismatch for the guard's own jq -e read.
  - file: scripts/install-local.sh

**Tests Run**
- git diff 4338fe9b068f58745db865e1c196f7b2162768fe..7d09f1f36685120939fcdbd254b39e149fd80da8 -- scripts/install-local.sh
- dash extract.sh valid-existing.json 0  (golden path, existing config) -> exit=0, phrases/fuzz preserved, full merged output
- dash extract.sh fresh-init.json 1  (fresh install) -> exit=0, default phrases/fuzz applied
- dash extract.sh corrupted.json 0  (unparseable config) -> exit=5, no install reached, matches pre-PR baseline behavior
- dash extract.sh empty-phrases.json 0  (valid JSON, phrases:[]) -> exit=1, guard refuses, no output left
- dash extract.sh external-detector.json 0  (wake.detector=external, phrases:[], command set - Validate()-legal) -> exit=1, guard wrongly refuses (P2 finding)
- bash extract.sh valid-existing.json 0 / fresh-init.json 1 / corrupted.json 0 -> identical results to dash, confirming quoting/composition is correct in both shells
- dash -n scripts/install-local.sh; bash -n scripts/install-local.sh -> both syntax OK
- gofmt -l . -> clean, no output
- CGO_ENABLED=0 go vet ./... -> exit 0 (plain `go vet` failed only due to a sandbox permission error reading /usr/include/stdc-predef.h for cgo, unrelated to this diff)
- which shellcheck / apt-get install shellcheck -> unavailable, no network/apt access in this environment; shellcheck was NOT run (static_only for that specific check)

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d66fc954ad3ccf-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/voice#36 at head c8f5cd5, diffed against the last-seen head 7d09f1f. The only change is replacing the hand-written jq wake.phrases guard with a call to the real binary (`voice --config FILE config`, which runs config.Load -> Validate). Built the actual binary and exercised it end-to-end (dash and bash) against the exact guard+install logic: P1 (pipe-truncation) still holds - a corrupted config aborts at jq under set -eu before the destructive install, in both shells. P2 (external-detector-with-no-phrases wrongly refused) is now RESOLVED - the binary-backed guard accepts that config (verified) and still correctly rejects an stt-detector config with empty phrases (verified, using the binary's own Validate error text). P3 (predictable /tmp path) is unchanged, still open, non-blocking. One new non-blocking defect found: the `config` subcommand used as validator is not side-effect-free - it runs session.Build, which (with the default Timers.Enabled=true) creates a SQLite timers.db under the invoking owner's XDG state dir as a side effect of merely 'printing effective config', reproduced by deleting the file and re-running the exact guard invocation. This doesn't threaten config-loss correctness but supports the requester's own instinct that a dedicated `validate` subcommand (config.Load+Validate only, no session.Build) would be cleaner than reusing the print command. Bootstrap ordering is safe: go build (line 18) happens under set -eu before /usr/local/bin/voice is installed (line 26), which is itself before the guard (line 98), so there is no window where the guard runs against a binary that doesn't exist; a compiling-but-buggy binary that rejected all configs would fail loudly at the guard (not silently) with the original /etc/voice/config.json left untouched, and would in any case have failed the later restart/hash-verification step regardless of this diff. The guard's own TOCTOU exposure (unprivileged check then privileged install of the same /tmp 
[truncated]
**Severity:** `P3`

**Findings**
- **Pipe-truncation regression fix still holds at c8f5cd5** (P1)
  - continues_uid: jerryfane/voice#36-f1
  - detail: The single composed jq invocation (scripts/install-local.sh:85) and the abort-on-corruption behaviour it relies on are unchanged by the c8f5cd5 diff, which only touches the guard immediately below it (what happens to a config that DID parse). Re-ran the corruption case at this exact head under both dash and bash: jq exits 5 on an unparseable /etc/voice/config.json, set -eu aborts the script at that line, and the destructive `sudo install ... /etc/voice/config.json` (line 104) is never reached, leaving the original file untouched. No regression.
  - evidence_locator: scripts/install-local.sh:70-85 (unchanged since 7d09f1f) and :104 (destructive step, unreached); commands: `dash extract.sh corrupted-config.json 0` -> exit=5; `bash extract.sh corrupted-config.json 0` -> exit=5, identical
  - file: scripts/install-local.sh
  - state: answered
- **Resolved: guard now defers to config.Validate via the real binary, and accepts wake.detector=="external" with empty wake.phrases** (P2)
  - continues_uid: jerryfane/voice#36-f2
  - detail: At c8f5cd5 the hand-written jq predicate `.wake.phrases | length > 0` was replaced with `/usr/local/bin/voice --config /tmp/voice-config.json config`, which runs config.Load -> Validate inside the actual application. Reproduced the exact scenario this finding named: a config with wake.detector="external", a non-empty wake.command, and empty wake.phrases now PASSES the guard (exit 0) under both dash and bash, where at 7d09f1f it was refused. Also confirmed the guard still correctly refuses the case that SHOULD fail - wake.detector="stt" (the default) with empty wake.phrases - reproducing the binary's own Validate error text ('wake.phrases is empty: Voice would never wake') rather than a second, independently-authored shell rule. This is the correct fix: it eliminates the duplicated-rule drift by making the application's own validator authoritative.
  - evidence_locator: scripts/install-local.sh:98 (`/usr/local/bin/voice --config /tmp/voice-config.json config`); internal/config/config.go:402-411 (Validate, unchanged); cmd/voice/main.go:104-107,147-153 (config subcommand calls config.Load then Validate before printing); commands: `dash extract.sh external-config.json 0` -> exit=0, detector=external, phrases=[]; `bash extract.sh external-config.json 0` -> identical; `dash extract.sh bad-stt-config.json 0` -> exit=1, binary's own Validate error surfaced
  - file: scripts/install-local.sh
  - state: answered
- **Intermediate /tmp/voice-config.json path is still fixed and predictable (unchanged by this diff)** (P3)
  - continues_uid: jerryfane/voice#36-f3
  - detail: The c8f5cd5 diff only changes what validates the file at /tmp/voice-config.json, not where that file lives or how it is created (`sudo jq ... > /tmp/voice-config.json`, unprivileged redirect, unchanged from 7d09f1f). Re-checked the requester's specific TOCTOU question this round: the guard's unprivileged check now execs the binary against the same fixed path instead of jq, which widens the window slightly (an extra process fork/exec plus config.Load/Validate/session.Build work) between the check (line 98) and the later `sudo install ... /etc/voice/config.json` (line 104) that re-reads the same path as root. However, /tmp is created with the sticky bit (mode 1777) on every standard Linux distribution, which prevents a non-owning local user from unlinking or renaming a file they don't own even in a world-writable directory - so a same-path swap attack by another local user remains blocked exactly as it was at the prior head. This is a pre-existing, non-worsened exposure; still worth a randomized/`mktemp`-based path in a future change, but it is not a defect newly introduced by this diff and does not block.
  - evidence_locator: scripts/install-local.sh:85 (unprivileged redirect, unchanged) and :98 (now execs the binary, widened but still sticky-bit-protected window) and :104 (root re-reads the same path); reasoning verified against standard /tmp semantics (mode 1777, sticky bit) rather than a live exploit attempt, since mounting a second unprivileged user account was out of scope for this review pass
  - file: scripts/install-local.sh
  - state: answered
- **Using `config` as the validator has a real side effect beyond validation: it creates a SQLite timers.db in the invoking owner's home directory** (P3)
  - detail: This directly answers the requester's second question (whether `config` is the right subcommand to reuse as a validator, versus a dedicated `validate`). `config` calls config.Load then session.Build(cfg) (cmd/voice/main.go:104-109,147-153) before marshalling and printing. session.Build (internal/session/build.go) unconditionally wires a timer scheduler when c.Timers.Enabled is true (the default), which calls timer.Open (internal/timer/store.go:45-63): 'Open prepares the database, creating the file and schema if needed.' Because the guard invocation is NOT run under sudo (scripts/install-local.sh:98), this runs as the unprivileged $owner, and Timers.File defaults to "" -> config.StatePath resolves to $XDG_STATE_HOME/voice/timers.db or ~/.local/state/voice/timers.db - i.e. a file gets created in the person-running-the-installer's own home directory purely as a side effect of a step whose comment describes it as a read-only check ('Belt and braces... refuse to install a config the binary itself will not accept'). Reproduced directly: deleted the state directory, ran the exact guard command in isolation, and the 8KB timers.db + parent directory were recreated. This is not destructive and self-heals (idempotent, harmless empty schema), and it does not threaten the config-preservation guarantee that is the point of this script, so it does not block merge - but it means the validator is not actually side-effect-free, which is precisely the failure mode the requester was probing for when asking whether `config` degrading later would be undetected. A dedicated `validate` subcommand doing only config.Load+Validate (skipping session.Build) would remove this side effect and also address the requester's own point about `config` becoming a weaker check if it were ever made more tolerant.
  - evidence_locator: scripts/install-local.sh:98 (guard invocation, unprivileged, no sudo); cmd/voice/main.go:104-109 (config subcommand calls session.Build unconditionally); internal/session/build.go:60-77 (Timers.Enabled branch opens a timer.Store); internal/timer/store.go:45-53 (Open creates the file/dir/schema if missing); reproduction: `rm -rf $XDG_STATE_HOME/voice; ./voice-test --config ./existing-config.json config > /dev/null; ls -la $XDG_STATE_HOME/voice` -> timers.db (8192 bytes) freshly created by that single invocation
  - file: scripts/install-local.sh

**Tests Run**
- CGO_ENABLED=0 go vet ./... -> exit 0, no output
- CGO_ENABLED=0 go build -trimpath -ldflags='-s -w' -o $TMPDIR/review/voice-test ./cmd/voice -> exit 0
- CGO_ENABLED=0 go test ./... -> all packages ok (cmd/voice, internal/audio, internal/brain, internal/codecheck, internal/device, internal/feedback, internal/hid, internal/session, internal/speech, internal/timer, internal/vad, internal/wake; config/faults/proc have no test files)
- gofmt -l cmd/voice internal -> no output (clean)
- which shellcheck -> not found on this host
- dash extract.sh corrupted-config.json 0 (extract.sh = unmodified guard/config_filter logic, lines 63-102 of scripts/install-local.sh at c8f5cd5, parameterized only on config path/fresh flag) -> jq parse error, exit=5, no install-equivalent step reached
- bash extract.sh corrupted-config.json 0 -> identical: jq parse error, exit=5
- dash extract.sh existing-config.json 0 (wake.phrases=["hey boys","hey voice"], fuzz=0.34) -> exit=0, GUARD PASSED, phrases/fuzz preserved unchanged
- bash extract.sh existing-config.json 0 -> identical
- bash extract.sh config.json 1 (fresh_config=1) -> exit=0, phrases=["hey voice"], fuzz=0 (documented defaults)
- dash extract.sh external-config.json 0 (wake.detector=external, wake.command=[...], wake.phrases=[]) -> exit=0, GUARD PASSED, detector=external, phrases=[] (P2 resolved)
- bash extract.sh external-config.json 0 -> identical
- dash extract.sh bad-stt-config.json 0 (wake.detector=stt (default), wake.phrases=[]) -> exit=1, 'refusing to install a config the voice binary rejects', underlying binary error 'wake.phrases is empty: Voice would never wake', /tmp-equivalent file removed
- dash -c 'set -eu; false; echo SHOULD_NOT_PRINT' -> exit 1, nothing after false printed, confirming set -eu aborts a failing simple (non-piped) command immediately, which is what makes the build-then-install-then-guard ordering safe
- rm -rf $XDG_STATE_HOME/voice; ./voice-test --config ./existing-config.json config > /dev/null; ls $XDG_STATE_HOME/voice -> timers.db (8192 bytes) recreated by the guard-equivalent invocation alone, confirming the validator has a disk-writing side effect beyond validation
**Diagnostics:** workflow advance failed: review approval held: approval is bound to head c8f5cd5fc9af8009e4ff4ad0b47e78eb70eb7a6e but the observed pull request head is 7d09f1f36685120939fcdbd254b39e149fd80da8; holding until the observed head catches up, because an approval at a genuinely superseded head must not clear changes_requested

PR #40, opened 2026-09-18
fix(config): find the installed config, and report which one was loaded

### LINKED ISSUE #39: config discovery: every documented command fails for the account that runs Voice
Every documented diagnostic command fails for the account that actually runs Voice.

From the device, immediately after a successful install:

```
$ sudo -u voice-agent /usr/local/bin/voice wake 'hey boys'
voice: load config: open /home/voice-agent/.config/voice/config.json: no such file or directory (run `voice init`)

$ sudo -u voice-agent /usr/local/bin/voice doctor
voice: load config: open /home/voice-agent/.config/voice/config.json: no such file or directory (run `voice init`)
```

Both work once `--config /etc/voice/config.json` is passed by hand.

## Cause

`config.Path()` resolves a per-user path. `packaging/voice.service` passes `--config /etc/voice/config.json` explicitly, so the **service** is fine — but nothing else is. The install puts the config at `/etc/voice/config.json`, owned `root:voice-agent` mode 0640, and never tells the binary to look there by default.

So the config that the device runs from is invisible to every command a person would type to inspect the device.

## Why it is worth fixing rather than documenting

The suggested remedy is actively wrong: `run \`voice init\`` would write a **second** config into the agent's home, which the service does not read. Anyone following that advice ends up diagnosing a file that has no effect on the running assistant — and it would look like it worked.

This also undercuts the `requests` reporting merged in #34 and the `wake` command added in #36. Both exist to be run by a human after a deploy, and both need the path spelled out to be useful at all.

## Fix

Fall back to `/etc/voice/config.json` when no per-user config exists, before failing. Order: `--config` flag, then `$VOICE_CONFIG`, then the per-user path, then the system path. Report which file was loaded in `voice doctor`, so "which config am I looking at" is answered rather than assumed.

## Acceptance criteria

1. `sudo -u voice-agent voice doctor` and `voice wake 'hey boys'` work with **no** `--config`, reading the same file the service reads.
2. A per-user config still wins over the system one, so a developer's own config is not shadowed by a system install.
3. `voice doctor` prints the path it loaded.
4. The error when nothing exists anywhere must not suggest `voice init` in a way that creates a file the service ignores — it should name the system path it looked for.
5. The regression check must fail on today's code and must distinguish *which* file was loaded, not merely that loading succeeded. Two configs with different wake phrases make that observable; a single config proves nothing.

## Found by

The deploy receipt for `abf1cef`. pi-burj ran the commands exactly as I specified them, they failed, and rather than quietly adding `--config` and reporting success it reported both the failure and the working form. That is the sixth defect that seat has surfaced by refusing to paper over a gap.

jerryfane:
Closes #39. Found by the deploy receipt for `abf1cef`, where the commands I specified failed as written.

## The defect

```
$ sudo -u voice-agent /usr/local/bin/voice doctor
voice: load config: open /home/voice-agent/.config/voice/config.json: no such file or directory (run `voice init`)
```

An installed device keeps its config at `/etc/voice/config.json`, and `voice.service` passes it explicitly — so the service works and **every command a person would type to inspect the device fails**.

### The advice was the worse half

`voice init` writes a *second* config into the caller's home that the service never reads. Someone following the program's own instruction would tune a file with no effect on the running assistant and conclude it worked — a machine teaching its operator a false model of itself.

It also undercut both things shipped today: the spoken-requests report (#34) and `voice wake` (#36) exist to be run by a person after exactly the kind of deploy just completed.

## The fix

Reading resolves in order: `--config`, `VOICE_CONFIG`, the caller's own per-user path, then the installed `/etc/voice/config.json`. The caller's own config still wins when it exists, so a developer's file is not shadowed by a system install.

When nothing exists anywhere, the error names **both** files it looked for:

```
no config found; looked in /home/pi/.config/voice/config.json and /etc/voice/config.json.
Run `voice init` to write /home/pi/.config/voice/config.json, or pass --config PATH
```

`voice init` is still suggested — but only where it is the right answer, instead of being appended to every load failure including ones for a config that existed and was merely elsewhere.

## doctor names the file, unconditionally

```
config       OK   /etc/voice/config.json
microphone   OK   ...
```

Printed on success too, because "which config am I actually running" is the question that was unanswerable. The loaded path was already being computed at the call site and thrown away on a `_ =` line — the answer existed and was discarded.

## Tests

Two configs holding **different wake phrases**, so the assertion is *which* file was loaded, not merely that loading succeeded — a single config passes whichever path the code picks.

| Mutant | Failure |
|---|---|
| never consult the installed path | `Locate() failed while an installed config exists` |
| installed config shadows the caller's own | `loaded [installed phrase] …; the caller's own config must win` |

The `systemPath` seam exists so the fallback can be tested on a host with no `/etc/voice` — otherwise that test only runs on a device, which is a check that cannot fail anywhere else. Production never reassigns it.

## Credit

pi-burj ran the receipt commands exactly as specified, they failed, and it **reported the failure** rather than quietly adding `--config` and calling it success. That is the sixth defect that seat has surfaced by refusing to paper over a gap, and the reason the deploy receipt is trustworthy at all.

## Verification

`test=0 vet=0 errcheck=0 gofmt clean`, statuses captured directly.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d671b6023b1df9-1`

**Decision:** `approved`
**Summary:** PR #40 (head 9e35deb7c67b0f4038be82fb951da29737971855) fixes issue #39 by adding config.Locate() (--config > VOICE_CONFIG > caller's own path > installed /etc/voice/config.json), a NotFoundError naming both files it checked, and an unconditional 'config <path>' line in `doctor`. Build/vet/test/gofmt/errcheck all pass under CGO_ENABLED=0 (cgo is blocked in this sandbox); two hand-written mutants (drop the systemPath fallback; invert precedence to let the installed config shadow the caller's own) both fail exactly the intended test, confirming the tests pin real behavior rather than wording. Live binary runs confirm 'config OK <path>' prints on success via --config, VOICE_CONFIG, and per-user discovery, and confirm the not-found error names both candidate files. voice.service always passes --config explicitly and only sets XDG_STATE_HOME, so the running service never calls Locate/Path and cannot diverge from a human operator's discovery under the shipped unit. Two non-blocking P3 gaps found: (1) when HOME is unset AND no installed config exists, Locate returns the raw os.UserHomeDir error instead of a NotFoundError naming both locations, silently dropping the already-checked systemPath from the message (pre-existing behavior for that branch, narrow real-world exposure); (2) the not-found error's 'voice init' advice only names the per-user path it would write, not the /etc/voice/config.json path an operator on a device most likely wants, though it does list both files it looked in and a generic --config escape hatch. Same-model-family caveat: this review was performed by Claude Sonnet 5; the commit author is recorded generically with no disclosed implementing model, so if the implementing agent was also Claude-family this is a same-family review pending independent corroboration. No prior verdicts exist for PR #40 to reconcile.

**Findings**
- **P3**
  - description: When Path() fails (e.g. HOME unset) and systemPath also doesn't exist, Locate returns the raw os.UserHomeDir wrapped error instead of a NotFoundError naming both files, even though systemPath was already checked and appended to `looked`. Reproduced live with an ephemeral probe test (t.Setenv HOME=""); pre-existing behavior for this branch, narrow exposure since the shipped service unit and normal shells set HOME.
  - id: home-unset-error-degrades
  - location: internal/config/config.go:362-381
- **P3**
  - description: NotFoundError's advice ('Run `voice init` to write <own-path>, or pass --config PATH') is accurate but only names the per-user file init would write by default; it doesn't explicitly surface `voice init --config /etc/voice/config.json` as the action a device operator most likely wants, though both candidate files are named in the 'looked in' clause.
  - id: notfound-advice-favors-developer-path
  - location: internal/config/config.go:346-350; cmd/voice/main.go:191-198

**Tests Run**
- CGO_ENABLED=0 go build ./... (exit 0)
- CGO_ENABLED=0 go vet ./... (exit 0)
- CGO_ENABLED=0 go test ./... (all 16 packages ok, config: ok 0.016s)
- gofmt -l . (empty/clean)
- CGO_ENABLED=0 go run github.com/kisielk/errcheck@v1.9.0 -ignoretests -blank -excludeonly ./... (exit 0, no output)
- CGO_ENABLED=0 go test ./internal/config/... -run TestLocate -v (4/4 PASS: TestLocateFallsBackToTheInstalledConfig, TestLocatePrefersTheCallersOwnConfig, TestLocateHonoursAnExplicitOverrideEvenWhenAbsent, TestLocateErrorNamesEveryFileItLookedFor)
- Mutation test 1: removed the systemPath fallback branch in a scratch copy -> TestLocateFallsBackToTheInstalledConfig FAILs as expected
- Mutation test 2: inverted precedence (check systemPath before caller's own) in a scratch copy -> TestLocatePrefersTheCallersOwnConfig FAILs as expected
- Built cmd/voice binary and ran `voice doctor` via explicit --config, via VOICE_CONFIG, and via per-user HOME/XDG_CONFIG_HOME discovery -> all three print 'config OK <path>' as first line
- Ran `voice doctor` with nothing resolvable anywhere -> NotFoundError text names both the own-path and /etc/voice/config.json
- Ephemeral probe test with HOME unset and no installed config -> Locate() returns raw 'cannot locate home directory: $HOME is not defined' instead of NotFoundError naming both files

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d673bd02517c5f-1`

**Decision:** `approved`
**Summary:** Diffed 9e35deb..0fd9edf (internal/config/config.go, internal/config/locate_test.go) at exact head 0fd9edf. Both prior P3 findings are fixed: Locate() now always returns *NotFoundError (never the raw os.UserHomeDir error) with UserPath left empty when Path() failed, and NotFoundError.Error() branches on UserPath to offer `voice init --config <System>` explicitly when the per-user file wasn't resolvable, or both actions when it was. Verified with real builds/tests/vet/errcheck/gofmt, all clean; main.go's only interaction with NotFoundError is opaque error wrapping so the new fields don't break it. Answering the requester's four questions: (1) TestLocateErrorNamesTheInstalledPathEvenWithNoHome checks behavior (type + path substring), fine as-is; TestNotFoundAdviceNamesTheDeviceAction's `voice init --config <systemPath>` substring is closer to wording but is defensible because the actual defect was 'the message doesn't hand the operator working command syntax' — checking for just the bare path would let a regression back to the old single-suggestion message pass (it still names both paths in 'looked in'), so I'd keep it rather than weaken it; if you want more rewrite-tolerance, assert on the pattern `--config <systemPath>` appearing anywhere in the message rather than requiring it immediately after 'voice init', which still pins the actionable flag+path pairing without pinning the surrounding prose. (2) No caller is misled by advising only the device action when UserPath is empty — Path() only fails when VOICE_CONFIG, XDG_CONFIG_HOME, and HOME/os.UserHomeDir are all unavailable, so no per-user path exists to suggest, and the generic '--config PATH' fallback is still present in that branch for anyone who wants a different location. (3) Both prior behaviors (installed-config fallback, own-config-wins) still pass unchanged: TestLocateFallsBackToTheInstalledConfig and TestLocatePrefersTheCallersOwnConfig both PASS. (4) Keeping UserPath and System as separate fields (vs. der
[truncated]

**Findings**
- **HOME-unset path now names the installed config instead of degrading to the raw home-dir error** (P3)
  - continues_uid: jerryfane/voice#40-f1
  - file: internal/config/config.go:388-399
  - state: withdrawn
  - withdraw_reason: Fixed at this head: Locate() no longer returns the raw os.UserHomeDir error. When Path() fails, `own` is cleared and Locate always returns &NotFoundError{Looked, UserPath: "", System: systemPath}, naming the already-checked installed path. Verified by `go test ./internal/config/... -run TestLocateErrorNamesTheInstalledPathEvenWithNoHome -v` (PASS) and by reading internal/config/config.go:388-399.
- **Advice now surfaces the device install command, not just the per-user one** (P3)
  - continues_uid: jerryfane/voice#40-f2
  - file: internal/config/config.go:355-368
  - state: withdrawn
  - withdraw_reason: Fixed at this head: NotFoundError.Error() now explicitly offers `voice init --config <System>` for a device install whenever System is the file to create, either alongside the per-user command (when UserPath is set) or alone (when it isn't). Verified by `go test ./internal/config/... -run TestNotFoundAdviceNamesTheDeviceAction -v` (PASS) and by reading internal/config/config.go:355-368.

**Tests Run**
- go vet ./internal/config/... -> exit 0
- CGO_ENABLED=0 go build ./... -> exit 0 (whole module, including cmd/voice which calls into this package)
- go test ./internal/config/... -run TestLocate -v -> 5/5 PASS, including TestLocateFallsBackToTheInstalledConfig, TestLocatePrefersTheCallersOwnConfig, TestLocateErrorNamesEveryFileItLookedFor, TestLocateErrorNamesTheInstalledPathEvenWithNoHome
- go test ./internal/config/... -v -> 9/9 PASS, including TestNotFoundAdviceNamesTheDeviceAction
- gofmt -l internal/config/config.go internal/config/locate_test.go -> no output (clean)
- go install github.com/kisielk/errcheck@latest && errcheck ./internal/config/... -> exit 0, no output

Issue #38, opened 2026-09-18
config: addressing audio by card number breaks silently when USB devices renumber


jerryfane:
The shipped config addresses the microphone by **card number**:

```
"device": "plughw:2,0"
```

Card numbers are assigned in probe order. They are not stable across a reboot or a reconnect, and plugging in any other USB audio device can shift them.

## Evidence this is live, not theoretical

A Creative Pebble V3 (`041e:3272`) was attached to the device today. Enumeration afterwards:

- Pebble: independent USB playback card, **card 4**
- PowerConf: **still card 2** — so nothing broke *this time*

That is luck, not design. Probe order put the new device after the existing one. Reverse the plug order, or reboot with the Pebble attached and the PowerConf briefly absent, and `plughw:2,0` points at the Pebble — a playback-only device with **no capture stream at all**.

## Why it matters more than a wrong device

The failure would not look like a configuration problem. Capture would fail or return silence, and every symptom would point at the software: the wake word never fires, `voice doctor` reports a microphone problem, and the obvious suspects are the VAD and the wake matcher. This repo has already spent effort on a capture reading that turned out to be the PowerConf cancelling its own speaker (issue #9); a silently renumbered card is the same kind of time sink with a cheaper fix.

## The fix

ALSA exposes stable named targets. From the device:

```
plughw:CARD=V3,DEV=0        # the Pebble
```

The PowerConf has an equivalent `CARD=` name. The defaults, the installer and the documented examples should use the **named** form, and `voice doctor` should report the resolved card name next to the configured string so a mismatch is visible before it is mysterious.

## Acceptance criteria

1. Default and documented device strings use `CARD=<name>` rather than a bare index.
2. `voice doctor` prints the configured device *and* the card name it resolved to, so "it says card 2 and card 2 is the wrong device" is readable rather than deduced.
3. A device string naming a card that does not exist fails with the available names listed, not with a generic open error.
4. The regression check must distinguish a renumbering from a working config. A test that passes with the device absent proves the code path was never exercised — the check must fail when a name resolves to a different card than intended.

## Not a deploy blocker

The current numbering is correct on the device as it stands, verified today, so this does not hold the pending install. It is the next thing after it.

PR #41, opened 2026-09-18
feat(spotify): play to the Pebble, with no path to the microphone

jerryfane:
The owner wants music on the speaker he just attached. The client must not be able to touch the microphone the assistant holds — and none of this may live only on the device.

## Why it is in the repo at all

A long-lived thing that exists because somebody typed it once on a host is a defect with a delay fuse. This fleet lost three such things today. So the ALSA namespace, the udev rule, the wrapper, the unit and the installer are all here, and a reimage reproduces them.

## The client cannot name its own output device

`spotify-player` calls `audio_backend::find(None)` and then `backend(None, ...)`. Librespot's ALSA backend maps that `None` to the literal string `"default"`. There is no device setting to pin — so the device cannot be chosen by configuring the client.

`packaging/spotify/alsa.conf`, selected through `ALSA_CONFIG_PATH`, therefore defines what `"default"` **means** for that process:

```
pcm.pebble { type hw; card "V3"; device 0 }
pcm.!default { type plug; slave.pcm "pebble" }
```

By **name**. Not `plughw:4,0` — card numbers are assigned in probe order, #38 exists for exactly that, and hardcoding a number I have just documented as unstable would be writing the next defect deliberately.

## That alone is advice, so the boundary is separate

An ALSA config is something a client could override. The account is something it cannot.

- `spotify-player` runs as its own account, **not** in the `audio` group
- a udev rule grants it the Pebble's nodes **only**, matched by USB vendor/product rather than card number
- every other card — including the capture device the Voice service holds — fails with `EACCES`

A config a client can ignore is advice; a permission it cannot escape is a boundary. The two are deliberately not the same mechanism.

## The installer proves its own pin is load-bearing

An ignored `ALSA_CONFIG_PATH` looks exactly like a working one: audio still plays, just out of the wrong speaker. So the installer points `default` at a nonexistent card and **requires playback to fail** before continuing.

### And it requires the failure to be attributable

My first version accepted any non-zero status. On a host without `alsa-utils` that passed because `aplay` did not exist — a check that could not fail, written into the check meant to prevent one. Caught by running it here:

```
$ ALSA_CONFIG_PATH=broken.conf aplay -d 1 /dev/zero
error: command not found: aplay      # "failed as required" — for the wrong reason
```

It now requires `aplay` to be present and the error to name the missing card, or it refuses to conclude anything.

## Not automated, deliberately

Login opens a browser flow against the owner's Spotify account. The installer prints the single command that needs a person and stops. Premium is required.

## Verification

`dash -n` clean on both installer scripts; Go suite unaffected (`test=0`). The attributable-failure guard was exercised on this host and correctly refused to proceed.

**Not verified here**: the install itself, the `EACCES` boundary and concurrent playback with Voice listening — all of which need the device and are next, with device-level evidence (`fuser /dev/snd/*`) rather than a listening test.


jerryfane:
> Agent: `voice-review-claude-temp-local-review-review-router-18d675254f75954c-1`
> Runtime: `claude`
> Job: `local-review-review-router-18d675254f75954c-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/voice PR #41 at exact head 81efd25c9847b1c8d18c398a778af0ac2688bc68 (3 commits: cc31a0c install+rule+conf, 280dba7 paired-control probe, 81efd25 ATTRS{../idVendor}→ATTRS{idVendor} fix) against base abf1cef. This is a non-Go diff (udev rule, alsa.conf, systemd unit, shell installer, README) touching zero .go files. The single prior recorded verdict for this PR (job local-review-review-router-18d6722a00f57283-1, head cc31a0c) is a daemon-crash 'failed' entry with zero findings — there is no substantive prior finding to answer. Independently re-derived and executed the operator's six checks: (1) over-grant — not reachable via hub reorder/card renumbering, because udev's ATTRS ancestor-walk binds to the specific USB device instance at add-event time, not to a numeric card slot, so a later device reusing that card number gets a fresh add event re-evaluated against its own (non-matching) ATTRS; same-model-family caveat genuinely applies (a second Pebble, or any hypothetical sibling product sharing USB PID 041e:3272, would be granted identically — stated per operator instruction). (2) ACL persists across suspend/resume (same inode, no add event) and is correctly reapplied on true replug (fresh add event); only a generic, inherent small race window between add-event processing and setfacl completion exists, common to any udev RUN+=-based ACL scheme, not a regression. (3) Paired-control logic in scripts/install-spotify.sh independently reproduced with three stub 'aplay' binaries (honored/ignored/broken) run directly (not through a pipeline): honored→VERIFIED(exit 0), ignored→IGNORED(exit 3, caught and refused), broken→UNPROVEN(exit 4, caught, does not conclude from the negative alone) — matches the operator's own three-stub result. Specifically addressed 'bogus fails for an unrelated reason': bogus.conf and the real alsa.conf share the same base include (/usr/share/alsa/alsa.conf), so a broken/missing base fails both arms symmetrically (UNPROVEN), not a fa
[truncated]

**Findings**
- **P3**
  - description: Over-grant analysis: ACL cannot land on an unrelated card after hub reorder/renumbering, since udev's plain ATTRS{} match walks the ancestor tree of the specific device instance at add-event time, not a cached numeric card slot; a later device reusing a freed card number triggers its own add event evaluated against its own (non-matching) ATTRS. Same-model-family caveat: any second device sharing USB PID 041e:3272 (a second Pebble, or a hypothetical mic-equipped sibling using the same PID) would be granted identically — not distinguishable by this rule.
  - evidence_locator: packaging/spotify/99-voice-spotify-pebble.rules:14-22 (static analysis of udev ATTRS ancestor-walk semantics); udevadm verify (syntax pass, no hardware available to run udevadm test)
- **P3**
  - description: Paired-control probe in scripts/install-spotify.sh independently reproduced with three stub aplay binaries (honored/ignored/broken), run directly with statuses captured (no pipeline): results were VERIFIED/IGNORED/UNPROVEN respectively, confirming the design cannot silently pass on a broken or config-ignoring host. The 'bogus fails for an unrelated reason' scenario cannot produce a false VERIFIED because bogus.conf and the real alsa.conf share the same base include (a broken/missing base fails both arms symmetrically into UNPROVEN) and because an ignored ALSA_CONFIG_PATH makes both arms consult the identical default namespace (forcing a symmetric result). Residual, low-probability risk: environmental non-determinism between the two sequential probes (e.g. a transient busy-device race) could in principle produce an asymmetric result unrelated to the namespace switch — an inherent limitation of any two-shot empirical test, not a design defect.
  - evidence_locator: scripts/install-spotify.sh:69-101; independent harness executed in this session against 3 stub aplay binaries (honored/ignored/broken), exits 0/3/4 respectively
- **P3**
  - description: DeviceAllow=char-alsa rw in the systemd unit is a coarse, class-wide cgroup allow (all ALSA char devices), not a per-node restriction; the Pebble-only boundary rests entirely on the udev-granted ACL plus the account's exclusion from the audio group (Unix DAC), not on this systemd directive. This matches the design's own stated model but is worth being explicit that DeviceAllow itself adds no card-level restriction.
  - evidence_locator: packaging/spotify/spotify-player.service:23-24 vs packaging/spotify/99-voice-spotify-pebble.rules:14-22
- **P3**
  - description: No hardware (Pebble/PowerConf USB devices) was available in this sandbox (/sys/class was Permission denied), so the specific hardware-attribute claims (real idVendor/idProduct values, and the ALSA card ID string 'V3' resolving correctly against the real device) could not be independently re-executed against hardware; relied on udevadm verify (syntax pass) and static semantic review. This is consistent with, and does not contradict, the PR's own disclosure that the Spotify install, EACCES boundary, and concurrent-playback checks remain open post-merge.
  - evidence_locator: packaging/spotify/99-voice-spotify-pebble.rules; packaging/spotify/alsa.conf:14-18; udevadm verify run in this session (pass, syntax-only)

**Tests Run**
- udevadm verify packaging/spotify/99-voice-spotify-pebble.rules — 1 file checked, Success: 1, Fail: 0
- sh -n scripts/install-spotify.sh — syntax OK
- git show 81efd25 (full diff of the head commit) — confirms ATTRS{../idVendor}/ATTRS{../idProduct} replaced with plain ATTRS{idVendor}/ATTRS{idProduct} on the pcm rule only; card rule (controlC%n) already used plain ATTRS and is unchanged
- Independent paired-control harness extracted from scripts/install-spotify.sh lines 69-101, run against 3 stub aplay binaries in $HOME/pairtest: honored stub → 'RESULT: VERIFIED', exit 0; ignored stub → 'RESULT: IGNORED', exit 3; broken stub → 'RESULT: UNPROVEN', exit 4 (both arms' captured output shown, not concluded from the negative alone)
- go build ./... — pure-Go packages unaffected (no .go files in this diff); cgo-importing packages fail with 'fatal error: /usr/include/stdc-predef.h: Permission denied', a sandbox environment limitation unrelated to this PR
- go vet ./... — same cgo-related environment failure, non-cgo packages unaffected
- go test ./... — internal/audio, internal/codecheck, internal/config, internal/feedback, internal/hid, internal/requests, internal/speech, internal/vad, internal/wake all pass ('ok'); cmd/voice, internal/brain, internal/device, internal/session, internal/timer fail to build due to the same pre-existing cgo/sandbox permission issue, not this diff
- git diff abf1cef..81efd25 --stat and per-file diffs reviewed in full (README.md, packaging/spotify/*.rules|conf|service|spotify-player-run, scripts/install-spotify.sh)

PR #42, opened 2026-09-18
fix(spotify): make the ALSA pin load-bearing by not including the base config

jerryfane:
The paired control in #41 caught this on the device before anything was installed. That is the only reason this is a commit rather than the owner hearing music from the wrong speaker.

## The pin was decorative

`packaging/spotify/alsa.conf` included `/usr/share/alsa/alsa.conf`. That file ends with an `@hooks` section loading `/etc/alsa/conf.d/*`, and **ALSA runs hooks after the root config is parsed** — so the PipeWire file dropped there redefines `pcm.!default`, replacing an override written *before* the hooks fire.

Playback therefore worked, out of whatever the desktop stack chose, and the named card was never consulted.

## Measured, not reasoned about

With the include, the **negative arm passed** — playback succeeded against a config naming a nonexistent card — and the installer refused to continue:

```
ALSA_CONFIG_PATH is being IGNORED: playback succeeded against a config naming a nonexistent card.
The Pebble pin would be decorative, so refusing to continue.
exit 1
```

Confirmed on the device: `default` resolved to `sysdefault`/`default` from the desktop stack, and `aplay -L` under the installed file showed the PulseAudio hook winning.

Standalone, statuses captured directly:

| namespace | result |
|---|---|
| bogus card | `aplay: audio open error: No such device`, `bogus_exit=1` |
| real card | `real_exit=0`, audible from the Pebble |

`hw` and `plug` are built into libasound, so the base config was never needed — and without it, nothing loaded later can override this default.

## What this says about my verification

My check on this file was a three-stub simulation of the control *logic*, which was sound, plus an **untested assumption about ALSA**. The logic test passed and the assumption was wrong. The check I built to catch exactly this class of error is what caught my own instance of it — which is the argument for building it into the installer rather than leaving it as a step someone performs.

## Independently confirmed at the same time

The udev rule **did** match on install:

```
$ getfacl /dev/snd/pcmC4D0p
user:spotify-player:rw-
```

So the permission boundary is real and was never the problem. The two mechanisms being separate is why one failing did not silently take the other with it.

## Verification

`dash -n` clean; no Go changes. The device re-runs the installer from merged `main` — nothing hand-edited there.

**Not verified yet**: the client build, the `EACCES` refusal on the PowerConf capture node, and Voice answering while music plays. Those resume once this lands, since the installer gate correctly blocked them.


jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d6769eb346b831-1`

**Decision:** `failed`
**Summary:** gitmoot_result severity is required when a review requests changes
**Diagnostics:** gitmoot_result severity is required when a review requests changes

jerryfane:
> Runtime: `claude`
> Job: `local-review-review-router-18d67740be60b01a-1`

**Decision:** `approved`
**Summary:** Independent re-review of jerryfane/voice PR #42 at head c76de5811691d30504a76a71ddbb1a4fe7c7338f (base a075efe, 2 commits, 3 files: README.md, packaging/spotify/alsa.conf, scripts/install-spotify.sh). No prior verdict for PR #42 was found in the ledger (checked GITMOOT_PRIOR_VERDICTS, 53 entries, none for PR 42), consistent with the requester's account that the earlier verdict at 97e2215 hit a delivery_error and never landed; I re-derived the finding independently from the diff rather than trusting the operator's paraphrase. The claimed P1 fix is real and verified: `diff <(sed -n '83,90p' scripts/install-spotify.sh) <(sed -n '30,37p' packaging/spotify/alsa.conf)` shows the two pcm.!default blocks differ in exactly one line, `card "NoSuchCardExists"` vs `card "V3"` — both arms are now `type plug` over `type hw`/`device 0`, self-contained, no base-config include. Q1 (comparability): no permission asymmetry — the real file is installed `sudo install -m 0644` (root-owned, world-readable) and the bogus file is written by the invoking user into their own mktemp dir; `probe_playback` runs `aplay` unprivileged (no sudo/spotify-player switch), so the same account reads both, closing the gap the operator worried about. One structural asymmetry does exist and predates this PR unchanged (confirmed via `git show a075efe:...`): the real alsa.conf carries a `ctl.!default { type hw; card "V3" }` block that bogus.conf lacks entirely, so 'the arms differ in exactly one line' is precisely true only for the pcm.!default block, not the full file. This is not a new defect and does not undermine the control, because aplay's plain playback path (no -M/mixer flag) never opens ctl.!default. Q2 (P3, librespot's actual ALSA usage): fetched librespot's dev-branch alsa.rs directly — `None` device maps to the literal string `"default"`, opened via `PCM::new(dev_name, Direction::Playback, false)`; namehint enumeration (`list_compatible_devices`) only runs when the device argument is literally `"?"
[truncated]

**Findings**
- **P3**
  - description: bogus.conf (scripts/install-spotify.sh) omits the ctl.!default block that the real alsa.conf carries, so the paired control's two files are not byte-for-byte identical except for the card name — only the pcm.!default stanza under actual test is. Pre-existing (unchanged since a075efe), and inert for this control since plain `aplay -q -d 1 -f cd /dev/zero` never touches ctl.!default. Non-blocking; noted for the record since the operator's 'exactly one line' claim is precise for the tested stanza but not the full files.
  - locator: scripts/install-spotify.sh:82-91 vs packaging/spotify/alsa.conf:30-42

**Tests Run**
- dash -n scripts/install-spotify.sh -> exit 0, clean
- diff <(sed -n '83,90p' scripts/install-spotify.sh) <(sed -n '30,37p' packaging/spotify/alsa.conf) -> single-line diff, card name only
- grep -rn 'usr/share/alsa/alsa.conf' and 'ALSA_CONFIG_PATH' across repo -> only historical/explanatory prose remains, no active stale include
- git show a075efe:packaging/spotify/alsa.conf and git show a075efe:scripts/install-spotify.sh -> confirms ctl.!default and the install/cargo/udevadm/authenticate scaffolding predate and are untouched by this PR
- CGO_ENABLED=0 go test ./... -> all packages ok
- gofmt -l . -> empty (clean)
- WebFetch https://raw.githubusercontent.com/librespot-org/librespot/dev/playback/src/audio_backend/alsa.rs -> confirms None device maps to literal "default", PCM::new(dev_name,...), namehint enumeration gated behind literal "?" only, not normal playback

